PR #3, opened 2016-08-04
Outdated link

lipis:


bennyn:
Maybe we should link to http://electron.atom.io/?

lipis:
and just Electron :) No need for GitHub's :D

Issue #6, opened 2016-08-10
Linux starts in full screen


ConorIA:
Hi folks, I know you are not officially supporting Linux at the moment, but I figured I'd put this on your radar anyway. When the app is built on Linux, the login screen starts in a window on first run. However, subsequent runs of the app will launch in full-screen mode with no menu bar, and no way of returning to windowed mode. I'm happy to try and troubleshoot. 

On an unrelated note, I'm maintaining an AUR PKGBUILD over here: https://aur.archlinux.org/packages/wire-desktop/

fenrig:
I second these on 2 different arch linux devices, needs to be fixed asap

fenrig:
I think it has something to do with line 148 in "electron/main.js" but I dont have any experience with electron whatsoever

lipis:
Fixed in #18 :) Thanks playing with it..

ConorIA:
Cheers!

fenrig:
Thank you!

lipis:
![penguins-aqQsmtq6J2L16](https://camo.githubusercontent.com/b9c2c5bb168f15fc8fa35f160a62cf6c394c2a42/68747470733a2f2f6d65646961312e67697068792e636f6d2f6d656469612f617151736d7471364a324c31362f67697068792e676966>)

geileszeuch:
It definitely doesn't start in fullscreen anymore, but now I cannot see half of the screen. Is there some secret combination to go to fullscreen?
![1](https://cloud.githubusercontent.com/assets/1734974/17945967/7b63c108-6a46-11e6-98be-56e9a2db3ee4.jpg)

ConorIA:
Can you resize the window in the bottom right? I had no trouble changing the width in GNOME (there does seem to be a minimum height though).

geileszeuch:
I could not resize it before, but now it does work (even after reverting the change to utils.js). Must be something wrong with my system.

PR #15, opened 2016-08-21
Resolve fullscreen on Linux issue

ConorIA:
This is a rough hack to get around issue #6. I have the app detect if the platform is linux on startup, and if so, resize. This code probably addresses the symptoms of the problem, but doesn't address that root cause. I have *zero* experience in Electron, so this was mostly trial and error. 

I think the real issue is that the app is automatically launching in fullscreen in Linux, thus, when the app initializes and it undergoes the 'if smaller than' test, on line 150 of `electron/main.js` it isn't resized using `resizeToBig`.

### DIFF
--- a/electron/js/util.js
+++ b/electron/js/util.js
@@ -53,6 +53,16 @@ module.exports = {
     win.center();
   },
 
+  resizeLinux: function(win) {
+    win.setMenuBarVisibility(true);
+    win.setFullScreen(false);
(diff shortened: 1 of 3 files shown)


lipis:
Thank you @ConorIA for experimenting with our app and trying to fix the error. Since you already mentioned that this is a hack we tried to solve it with less changes in #18 :)

PR #24, opened 2016-08-30
Reloading the app when the render process crashes

bennyn:
Suggestion by: https://github.com/electron/electron/issues/6971#issuecomment-242803813

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -221,7 +221,7 @@ function showMainWindow() {
   });
 
   main.webContents.on('crashed', function() {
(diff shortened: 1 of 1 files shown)


lipis:
:D

bennyn:
Approved?

lipis:
how could we crash the webContents to simulate/test it :)

`process.crash()` I think crashes the whole thing..

lipis:
I tested it.. and it works.. white page and then reloads :) better than a real crash..

PR #33, opened 2016-09-07
Make tray icon a toggle

ConorIA:
This one is a little more presumptive and could go against your own design spec. 

The push turns the tray icon into a toggle:

If Wire is hidden, it is shown. 
If Wire is shown, but does not have focus (under other windows) it is shown. 
If Wire is in focus, it is hidden.

This is good for my personal workflow, but I will understand if you folks aren't into it! 👍


lipis:
We'll have to pass on those for various reasons! We are not officially support Linux and we have our lovely design team that does all the research and experiments! We'll take that into consideration though.. Thanks for the suggestion..

PR #65, opened 2016-10-11
Update amplify call for preferences

gregor:


gregor:
Merging this now as the new preferences have made it into staging

herrmannplatz:
Is this backwards compatible?

gregor:
Should be. The old call is still referenced in https://github.com/wireapp/wire-webapp/blob/dev/app/script/view_model/list/ListViewModel.coffee#L61. Can be removed once uptake of new version is high enough. Not sure how this is related to WEBAPP-3267

gregor:
Tried to keep backwards compatibility in mind with all the changes but apparently missed it in some places. We need to come up with a better strategy to keep track of the depencies between wire-desktop and wire-webapp moving forward.

Issue #75, opened 2016-10-12
Prevent Wire from running more than one instance


ElvisPresly:
Make please setup: To launch no more than one copy of the program.
For example. I use Linux and if Wire is already launched, and I click on an icon the program twice (not in a tray), then one more copy of the program is launched. It is always necessary to go to a tray and from there to tear the Cryptocat window. To tear a window from Dock or from a desktop there is no opportunity.
Thanks

ConorIA:
This has been addressed in #77.  :smile:

ConorIA:
@lipis, this can probably be marked as closed, unless you are keeping it open for something?

lipis:
@ConorIA thanks :) Just missed it..

PR #77, opened 2016-10-12
Make single instance stuff work on Linux too

ConorIA:
This should address #75. Is there any reason why you don't want to single instance stuff to work on Mac? (i.e. Why was it Windows-only?)

~~EDIT: Oops. Doesn't seem to have done the trick.~~ There was a command missing. Works now.


lipis:
The comment right above the if explains exactly that... and on macOS it's only one instance anyway.. that's how all apps are working :)

electron/main.js
```diff
@@ -86,6 +86,9 @@ if (process.platform === 'win32') {
   });
 }
 
+if (shouldQuit) {
+  app.quit();
+}
```
ConorIA: It seems to be working ok on my Manjaro box. What about something like: 

```
if (process.platform === 'linux' && shouldQuit) {
  app.quit();
}
```

Should that be a single `&` or double `&&` in JS?

lipis: I would also put it inside the above `if` but again.. we'll have to investigate that more.. for example if the app is minimized.. will it show the current instance that is running when you try to open a new one?

electron/main.js
```diff
@@ -84,9 +84,11 @@ if (process.platform === 'win32') {
     }
     return true;
   });
+  if (process.platform === 'linux' && shouldQuit) {
+    app.quit();
+  }
 }
 
-
```
lipis: leave that untouched :)

ConorIA: Sorry, I thought it was a line I had added by mistake!

lipis:
Thank you!

Issue #71, opened 2016-10-11
Create "File" menu for Linux version


lipis:
Because things like settings and stuff are under that menu :/

<img width="169" alt="screen shot 2016-10-11 at 23 00 40" src="https://cloud.githubusercontent.com/assets/125676/19288492/8ecb0a9a-9006-11e6-9f22-5fc2ab7f5c7a.png">

ConorIA:
I had no idea this was missing, but it certainly would greatly improve usability!

PR #83, opened 2016-10-13
Add Linux "Wire" menu, add fullscreen and menu bar toggles

ConorIA:
This is very, very basic, but I thought I'd take a shot at #71. All of the 4 items work.

menu

Note that I thought about adding the fullscreen options to the Linux menu as you've done for Mac, but I couldn't figure out how to get back to windowed mode without editing my config files.

### DIFF
--- a/electron/js/menu/system.js
+++ b/electron/js/menu/system.js
@@ -148,7 +148,7 @@ var toggleMenuTemplate = {
 var toggleFullScreenTemplate = {
   i18n: 'menuFullScreen',
   type: 'checkbox',
-  accelerator: 'Alt+Command+F',
+  accelerator: process.platform === 'darwin' ? 'Alt+Command+F' : 'F11',
   click: function() {
     getBrowserWindow().setFullScreen(!getBrowserWindow().isFullScreen());
   },
@@ -316,6 +316,13 @@ if (process.platform === 'linux') {
     i18n: 'menuPreferences',
     click: function() {sendAction('profile-settings-show');},
   });
+  windowTemplate.submenu.push(
+    separatorTemplate,
+    toggleMenuTemplate,
+    separatorTemplate,
+    toggleFullScreenTemplate
+  );
+  toggleFullScreenTemplate.checked = init.restore('fullscreen', false);
 }
 
 if (process.platform !== 'darwin') {
--- a/electron/js/menu/system.js
+++ b/electron/js/menu/system.js
@@ -268,7 +268,7 @@ var win32Template = {
 };
 
 var linuxTemplate = {
-  label: config.NAME,
+  i18n: 'menuFile',
   submenu: [
     localeTemplate,
     signOutTemplate, {
--- a/electron/js/menu/system.js
+++ b/electron/js/menu/system.js
@@ -267,6 +267,23 @@ var win32Template = {
   ],
 };
 
+var linuxTemplate = {
+  label: config.NAME,
+  submenu: [
+    {
+      i18n: 'menuPreferences',
+      accelerator: 'Ctrl+,',
+      click: function() {sendAction('profile-settings-show');},
+    },
+    localeTemplate,
+    signOutTemplate, {
+      i18n: 'menuQuit',
+      accelerator: 'Ctrl+Q',
+      click: function() {app.quit();},
+    },
+  ],
+};
+
 menuTemplate = [
   conversationTemplate,
   editTemplate,
@@ -296,6 +313,10 @@ if (process.platform === 'win32') {
   squirrel.startupLinkExists(function(exists) {
     menu.items[0].submenu.items[2].checked = exists;
   });
+}
+
+if (process.platform === 'linux') {
+  menuTemplate.unshift(linuxTemplate);
 }
 
 if (process.platform !== 'darwin') {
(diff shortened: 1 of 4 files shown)


ConorIA:
Just noticed something that isn't quite right: The quit menu item quits Wire, but the Ctrl-Q key combo just closes the window to the tray.

lipis:
on mobile now.. but on Linux is more common to have alt+F4 for quit.. no?

ConorIA:
Most apps use Ctrl-Q as the shortcut (I had a bunch of windows open anyway, so I tried it and it closed Nautilus, Atom, Thunderbird, Gedit, at least). Although Alt-F4 worked as well on Nautilus. 

Atom and Brave browser also use Ctrl-Q as the quit shortcut in menus. 

One thing that is certainly out of place in my example though is the preferences menu. Preferences are usually under "Edit" in Linux.

lipis:
Right.. I checked it as well.. `Ctrl+Q` should be OK. Preferences should go under edit though for consistency..

electron/js/menu/system.js
```diff
@@ -298,6 +315,10 @@ if (process.platform === 'win32') {
   });
 }
 
+if (process.platform === 'linux') {
+  menuTemplate.unshift(linuxTemplate);
```
lipis:
```js
  editTemplate.submenu.push(separatorTemplate, {
    i18n: 'menuPreferences',
    click: function() {sendAction('profile-settings-show');},
  });
```

lipis: something like that maybe..

ConorIA:
I'll make those changes tomorrow. Tonight Snarky Puppy is playing in Toronto! 😋

ConorIA:
There we go!

electron/js/menu/system.js
```diff
@@ -267,6 +267,18 @@ var win32Template = {
   ],
 };
 
+var linuxTemplate = {
+  label: config.NAME,
```
ConorIA: I guess so, though it isn't really a 'traditional' menu, is it? 
To change that, would line 271 change to `i18n: 'menuFile',`?

ConorIA: Oh no! I just pushed changes to make the menu 'File' (though you might need to double-check my German!). Anyway, on Monday let me know what the designers prefer and I can revert if necessary.

ConorIA:
Another quick observation from looking at other apps: Edit is virtually always beside File. 

That would mean that the Menus would be: 
Wire (or File), Edit, Conversation, Window, Help

I see where the menu order is specified in `system.js`, but I don't want to make any changes because it seems like that would change it for all apps. So I'll wait to hear from you whether that is a good thing to change or not. 

I'm not sure if it is just a coincidence, but most apps seem to use a hierarchy of: 

1) File, Edit, View
2) Application-specific menu items
3) Window, Help

electron/js/menu/system.js
ConorIA: Sounds good. I'm happy to tinker. I am learning a lot!

ConorIA:
In the end I decided to go ahead and push the changes from my last comment. It won't take much to revert them next week if necessary.

electron/js/menu/system.js
```diff
@@ -267,9 +267,21 @@ var win32Template = {
...
+      accelerator: 'Ctrl+Q',
+      click: function() {app.quit();},
+    },
+  ],
+};
+
 menuTemplate = [
-  conversationTemplate,
   editTemplate,
+  conversationTemplate,
```
lipis: rollback the order.. if anything could happen later in another review/pr

electron/js/menu/system.js
```diff
@@ -267,9 +267,21 @@ var win32Template = {
   ],
 };
 
+var linuxTemplate = {
+  label: config.NAME,
+  submenu: [
+    localeTemplate,
```
lipis: add a separator after locale

electron/js/menu/system.js
```diff
@@ -271,6 +271,7 @@ var linuxTemplate = {
   label: config.NAME,
   submenu: [
     localeTemplate,
```
lipis: So before the languages add Preferences and separator.. here as well (without removing the preferences from Edit)

electron/js/menu/system.js
```diff
@@ -298,6 +311,21 @@ if (process.platform === 'win32') {
...
 
+if (process.platform === 'linux') {
+  menuTemplate.unshift(linuxTemplate);
+  editTemplate.submenu.push(separatorTemplate, {
+    i18n: 'menuPreferences',
+    click: function() {sendAction('profile-settings-show');},
+  });
+  windowTemplate.submenu.push(
+    separatorTemplate,
+    toggleMenuTemplate,
```
lipis: remove the menu toggle from here.. doesn't seem to do much, there will be changes in Windows soon here so we'll rethink about it in the future

lipis:
Still here?

ConorIA:
Away from the computer! Won't be home until the evening. I think the option is enabled that should let you edit the PR, does that just mean edit the comments, or can you edit the commits?

lipis:
or we can take it from here... thank you! :)

ConorIA:
That works too! 😁

lipis:
You are right.. didn't think of that :) Next time.. thanks

PR #93, opened 2016-10-17
Linux menu items

lipis:


ConorIA:
Not sure if you'd rather I open an issue for this, but the preferences menu doesn't seem to be working on 2.11.2662. I can confirm that it is working on the build based on the PR that I sent. 

I also noticed that I forgot to include the `Ctrl+,` shortcut for preferences in the Linux menu.

PR #96, opened 2016-10-18
Avoid building every arch on Linux

ConorIA:
When building on Linux, the Gruntfile doesn't specify architecture, so we end up building on x64, i686 and ARM! 

Not sure if this will become redundant after the change to Electron-builder, but thought I'd send it your way any way. I've patched my PKGBUILD after a user complained about needing to download Electron four times. 😛

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -17,6 +17,7 @@
 #
 
 electron_packager = require 'electron-packager'
+os = require 'os'
 
 ELECTRON_PACKAGE_JSON = 'electron/package.json'
 PACKAGE_JSON = 'package.json'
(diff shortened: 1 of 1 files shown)


ConorIA:
Unrelated: is there any reason why the install process downloads both electron-v0.36.12 and electron-v1.4.3?

lipis:
Most likely we're going to get rid of this build in favor of electron-builder as you already mentioned..

ConorIA:
Sure thing. I played with electron-builder a little, though, and the "pacman"
build that it produces is nasty. We have way more flexibility with PKGBUILD.

>

PR #97, opened 2016-10-19
Added ES, FI, HR and RO

lipis:


lipis:
screen shot 2016-10-19 at 11 40 34

PR #98, opened 2016-10-19
Selected language

lipis:


gregor:
Can we be smarter about the language and default to the system language? We have the ticket WEBAPP-2594 to track that.

PR #116, opened 2016-10-26
Added Uzbek

Nuruddinjr:
Added Uzbek language Localization


lipis:
Hi @Nuruddinjr.. All the translations are being handled through Crowdin https://crowdin.com/project/wire-cs/uz feel free to translate them there and we will include it in the future releases :)

PR #133, opened 2016-11-01
Linux cleanup and deploy on S3

lipis:


ConorIA:
Please reconsider removing the option to build on Linux without electron-builder. That tool is imperfect for Arch-based distros. Sometimes one just wants to build for their own OS! This PR will break my PKGBUILD files, so I'll have to patch the 'linux' command back into `Gruntfile.coffee`.

lipis:
Fair point.. we will do something about it :)

ConorIA:
I got the issue solved with a little bit of prying. See comment on #138. 😄

PR #138, opened 2016-11-01
Replace ability to build locally on Linux

ConorIA:
This is probably not a perfect solution, but for now this is the patch that I am using for my PKGBUILDs, after the commands I was using were removed in #133.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -17,6 +17,7 @@
 #
 
 electron_packager = require 'electron-packager'
+os = require 'os'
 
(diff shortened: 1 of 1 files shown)


ConorIA:
This was nasty! I've come up with a much more elegant solution: 

```
npm install
grunt 'clean:linux' 'update-keys' 'release-prod'
if [ $CARCH == 'x86_64' ]; then
  node_modules/.bin/build --linux --x64 --dir
else
  node_modules/.bin/build --linux --ia32 --dir
fi
```

Issue #104, opened 2016-10-20
Preference to automatically launch


mcg:
It's possible to manually add Wire to auto-start, but it would be nice for the app to provide a preference to add this easily.

Finnerale:
A minimized auto-start would be cool too :+1:

ConorIA:
In the short-term, can you add the app to your startup applciations? I have Wire start automatically because it is included in my start-up programs, which I think is the easiest way to do it! Though a 'start minimized' option might be useful. 

As a bit of a case-study: Rambox is an app that does AutoLaunch via Electron. ([See](https://github.com/saenzramiro/rambox/search?utf8=%E2%9C%93&q=autolaunch&type=Code)). 
When I had the app installed, I thought it was really invaisive. I would disable the option, and remove it from my startup programs, but every time there was an update, the darn program would auto-launch again. This was my experience on Linux, but I didn't like it.

mcg:
My experience is a "Start on login" pref that is usually not checked. Once checked it adds itself to startup applications. Start minimized is also nice option.

ConorIA:
I've been working on the latter. :)

shvchk:
For now I'm using [this script](https://gist.github.com/shvchk/e46ce47efb721f2cbeb6d170e6215bac) in autostart (on Linux) to start Wire minimized.

Nateowami:
@shvchk Thanks! Is `--startup` an option specific to Wire, or is it an Electron thing? I don't see any documentation for it.

ConorIA:
@Nateowami it is specific to Wire. By the way, auto launch is coming in the next release!

PR #146, opened 2016-11-03
Add option to start minimized

ConorIA:
Adds a checkbox to the 'Window' menu to start the app in the tray (i.e. skip `main.show()` on startup.)
This partially addresses #104. 

I'm not sure what is going on in the Windows menu, so I didn't touch that, but I feel pretty confident about the code for macOS. Tried, tested and true on Linux, of course!


shvchk:
Would be nice to have this as an argument, like `--minimized`, so that Wire could be added to autostart to be launched minimized, and showed as usual if launched manually.

ConorIA:
I'm not entirely sure what the `--startup` flag is for, but if it is `true`, I think the app window won't show. It is on the same line as the change I made in this PR. @lipis, is this used anywhere else in the app?

ConorIA:
Come to think of it, my guess is that that flag is precisely for when you want Wire to launch on startup.  😌

shvchk:
Ha! There really is `if (!argv.startup)` check before `main.show()`, thanks for the tip!

ConorIA:
Thanks for the suggestion. It's addressed in the last couple of commits.

ConorIA:
Updated to include newer changes.

lipis:
Handled in #329

PR #148, opened 2016-11-03
Use yarn with Electron-Builder too

ConorIA:
With this commit, NPM can be removed as a dependency in favour of yarn. 

How important is the `npm rebuild` command before building? 

If it is imperative to rebuild the modules before packaging, a possible solution would be to add `yarn install --force` to the shell task in the Gruntfile.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -167,9 +167,9 @@ module.exports = (grunt) ->
 
     shell:
       linux64:
-        command: 'npm run linux64'
+        command: 'yarn run linux64'
       linux32:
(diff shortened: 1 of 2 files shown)


lipis:
Nice :)

ConorIA:
@lipis I'm worried I may have sold you out on this one. If any of the dependencies are architecture-specific, skipping the rebuild could cause the 32-bit build to fail...  You'd know better than me, anyway.

ConorIA:
Reference https://github.com/electron-userland/electron-builder/issues/861.

PR #154, opened 2016-11-04
Add option to auto-launch (#104)

ConorIA:
Addresses #104. I have currently only added the option to the Linux menu, pending feedback from Wire devs and designers. Cheers!

~~Edit: Travis is failing because he doesn't like the ternary operator.~~


lipis:
Interesting.. on Windows we have already the auto start as an option.. that is done manually :) We could check into that if it could be done in a more sophisticated way... we'll check next week..! Thanks

ConorIA:
Sure thing. 

Personally, I think that OS-level methods of autostart are usually sufficient, especially if the user wants to specify the `--startup` flag for instance. But, the issue reporter didn't agree, so I thought it'd be a good chance to teach myself how to use packages! 😉 

One thing to keep in mind is that the auto-launch module isn't compatible with MAS sandboxing. See here: https://github.com/Teamwork/node-auto-launch/issues/43

lipis:
It might cause issues with MAS version even if we're not going to use it.. since it will be part of the package :/ Right?

ConorIA:
I'm not sure. I don't know how the MAS submission process works. My guess would be that if we only provides the option to startup on Linux and Windows, they wouldn't have any reason to reject the app. In fact, we could even wrap the launcher code in an if statement. I think the only way to know for sure is to try it! 😜

ConorIA:
@adam-lynch, do you have any insight about whether we can use auto-launch on Linux and Windows and still keep things kosher for the MAS?

adam-lynch:
@ConorIA I asked a colleague who'd know more than me on this, specifically "If you had code in a desktop app which violated MAS rules but was never called, would it fail review?";

>potentially yes, it's a bit of a mystery how they check

I guess there's only one way to find out 😉 

---

Complete guessing from me... but maybe it might have more of a chance if it's obviously not going to be called? I.e. clearer code with less magic. Like maybe instead of `if(config.isMac)` have `if(false)` (replaced/injected at build-time). But that's a complete guess.

I suppose redundant code shouldn't ever be shipped (per platform), although we don't do that ourselves yet either.

ConorIA:
Thanks @adam-lynch. I appreciate your looking into this.

lipis:
Thanks guys.. we'll try to submit it disabled and see what happens.. if it doesn't produce many issues I'll be happy to drop the custom made solution that we have on Windows..

ConorIA:
Another caveat: This is probably incompatible with the AppImage, because the executable can be _anywhere_. 

This is the content of the `.desktop` when I tried it with AppImage: 

```
[Desktop Entry]
Type=Application
Version=1.0
Name=wire
Comment=wirestartup script
Exec=/tmp/.mount_NjYte7/usr/bin/wire
StartupNotify=false
Terminal=false
```

The `Exec` command disappears as soon as I close Wire, so the script won't work.

lipis:
It's getting hairy :(

ConorIA:
I've got an idea. I'll hack on it this afternoon. :)

lipis:
giphy

probonopd:
@ConorIA 

>Exec=/tmp/.mount_NjYte7/usr/bin/wire

You have to use `Exec=/path/to/Your.AppImage`

`/path/to/Your.AppImage` can be gotten programmatically at runtime by inspecting the `$APPIMAGE` environment variable.

probonopd:
> "Due to the portable nature of AppImage, we can\'t locate your executable. Please manually add Wire to your autostart items."

Why not edit `Exec=` on the fly to point to the value of `$APPIMAGE`?

ConorIA:
@probonopd, most of my programming for Wire is trial and error! I am happy to change what I have just implemented, but I'd need your hint as to where I can get the value for $APPIMAGE. Where does the AppImage locate get stored, and how can I call it into the script?

probonopd:
`$APPIMAGE` is an environment variable. I think you can access it at `process.env.APPIMAGE` from an Electron based app.

ConorIA:
@probonopd, I wish I had known about this before! That was way easier. Oh well. I guess I can say I learned something! :wink:

ConorIA:
@lipis, you were interested in trying to submit to the MAS before the AppImage issue. The problems with AppImage have since been resolved. Can we give it a shot?

lipis:
200w

I was just thinking about it today... I'll check it tomorrow thoroughly and most importantly with MAS.. sorry for the delay..!

electron/js/menu/system.js
```diff
@@ -226,6 +233,15 @@ var toggleFullScreenTemplate = {
   },
 };
 
+var toggleAutoLaunchTemplate = {
+  i18n: 'menuAutoStart',
+  type: 'checkbox',
+  checked: init.restore('shouldAutoLaunch'),
```
lipis: As we discussed in one other PR.. the `init.restore` is always used with the default value (you never know what can go wrong with the filesystem) so I would say use `false` wherever it is used..

electron/locale/strings.js
```diff
@@ -49,6 +49,7 @@ string.menuQuit = 'Quit Wire';
 string.menuShowHide = 'Show/Hide Menu';
 string.menuSavePictureAs = 'Save Picture As...';
 string.menuNoSuggestions = 'No Suggestions';
+string.menuAutoStart = 'Start with OS';
```
lipis: Since we have `string.menuStartup = 'Launch Wire on Windows Start';` maybe we can just remove Windows from that string and use the same for all.

electron/js/menu/system.js
```diff
@@ -28,6 +30,11 @@ const locale = require('./../../locale/locale');
 let menu;
 var menuTemplate;
 
+const launcher = new autoLaunch({
+  name: 'Wire',
```
lipis: just import `const pkg = require('./../../package.json');` and then use the `pkg.productName`

ConorIA: The tray menu uses config.Name. Should I use that or the pkg.Name as per your comment?

electron/js/menu/system.js
ConorIA: Is this correct? 
` checked: init.restore('shouldAutoLaunch', false),`

electron/locale/strings.js
ConorIA: Sure thing! Should I make the change in just `strings.js` or should I also change `strings-en.js`? I think the `string-en.js` is provided by Crowdin, right?

ConorIA:
I just noticed that I have `auto-launch` in both the dev deps and the prod deps. Is that correct? Or should it just be in `electron/package.json`?

electron/locale/strings.js
```diff
@@ -6,7 +6,7 @@ string.menuEdit = 'Edit';
 string.menuWindow = 'Window';
 string.menuView = 'View';
 string.menuHelp = 'Help';
-string.menuStartup = 'Launch Wire on Windows Start';
+string.menuStartup = 'Launch Wire on OS Start';
```
lipis: Remove the `OS` completely.. it's not that obvious for most..

ConorIA:
Finally, the template has only been added to Linux as of this PR. Should it be added to the Windows menus as well? Or will you keep your home grown startup solution?

lipis:
I was thinking about doing that afterwards as there are many things that needed cleanup! I'll merge it now and will test Windows afterwards. Thank you!

PR #164, opened 2016-11-08
Add license info to app package.json

ConorIA:
This will also define the license for the built apps.
Currently, the deb shows an 'unknown' license.

The LICENSE.md file from the repository root (which specifies GNU GPLv3) is not included in the final packages; it really should be. Instead, the Github license is included.

### DIFF
--- a/electron/package.json
+++ b/electron/package.json
@@ -7,6 +7,7 @@
   "updateWinUrl": "https://wire-app.wire.com/win/prod/",
   "environment": "production",
   "homepage": "https://wire.com",
+  "license": "GPL-3.0",
   "author": "Wire Swiss <wireapp@wire.com>",
   "private": true,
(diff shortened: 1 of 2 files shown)


lipis:
Update the main `package.json` the same as https://github.com/wireapp/wire-webapp/blob/dev/package.json#L3 while you are on it..

ConorIA:
Like that?

lipis:
yep.. :) thanks

ConorIA:
@lipis, the LICENSE.md that is distributed with the unpacked archive and with the deb folder is still the Github license from Electron. How can we include the GNU GPL-3 with the package?

lipis:
will check with legal actually if the product itself is under GPL-3.. or just a code :) you never know with these guys..

Issue #171, opened 2016-11-09
Update .deb dependencies


ConorIA:
I don't know for sure, but I imagine that the .deb has depencies outside of the default libappindicator stuff. On Arch, these are the depencies, as detected by pamac.

```
alsa-lib (alsa-lib-noassertion)
gconf (gconf-gtk2)
gtk2 (gtk2-aqd, gtk2-patched-filechooser-icon-view, gtk2-patched-gdkwin-nullcheck, gtk2-ubuntu)
libxss
libxtst
nss
```

Does Debian have a tool to detect dependencies? Or can a Debian user try running Wire in a clean chroot? 

I guess this isn't a huge issue, since no one has complained so far, but it might be worth it to have the metadata complete.

sonicou:
when installing wire on a minimal debian jessie system the following packages need to be installed:

libappindicator1
libnotify-bin

both of them will install a large list of dependencies themselfs (like gtk)

ConorIA:
I've installed the deb in a Docker container using the official ubuntu:latest Docker image. The following are missing dependencies: 

| Missing Shared Library | Package to Install |
|--------------------------------|--------------------------|
| libXss.so.1 | libxss1 |
| libgconf-2.so.4 | libgconf-2-4 |
| libnss3.so | libnss3 |
| libasound.so.2 | libasound2 |

sonicou:
I'm not so sure their docker image is a minimal ubuntu system or is it?

ConorIA:
Yes, very minimal. In fact, the missing deps might even be part of the default desktop install, but I figure it is best to be thorough.

lipis:
Solved in #188

PR #188, opened 2016-11-16
Update deb deps

ConorIA:
While these may not be missing from a typical Ubuntu install, Wire refuses to launch without at least these dependencies. This list generated from a the official Ubuntu Docker image (very minimal) in which I installed Wire. See #171 for the particular shared library `.so` files that were missing.


ConorIA:
Shout-out to sortmylist.com 😉

lipis:
Thanks!

PR #236, opened 2016-11-30
Webapp 3402

mythsunwind:
Only triggers platform-specific menu creation on triggering of createMenu function


lipis:
Better titles next time :)

PR #238, opened 2016-12-01
Compile linux for different architectures separately

mythsunwind:
The node module that does the spellcheck needs to be compiled for the architecture it is run. Currently we only compile it for the architecture of the build server, but we need need to compile it for the specific architectures (ia32 and x64) now.

Example: When compiling the linux build on x64 architecture, we currently build the spellcheck module for x64 architecture and package it together with either ia32 electron or x64 electron into deb files for ia32 and x64. The deb package for ia32 contains the ia32 electron but the x64 node module. Electron will potentially throw an error when trying to load the module, because it is not an ia32 module.

The module is build at the time `npm install` is run, therefor I propose having two separate tasks for each architecture. Maybe there is a better way.

### DIFF
--- a/electron/package.json
+++ b/electron/package.json
@@ -21,10 +21,10 @@
     "open-graph": "0.2.1",
     "raygun": "0.9.0",
     "request": "2.76.0",
-    "winston": "https://github.com/wireapp/winston.git#2.2.0-e",
-    "spellchecker": "3.3.1"
+    "spellchecker": "3.3.1",
+    "winston": "https://github.com/wireapp/winston.git#2.2.0-e"
   },
   "optionalDependencies": {
-    "node-addressbook": "https://github.com/wireapp/node-addressbook.git#1.0.1",
+    "node-addressbook": "https://github.com/wireapp/node-addressbook.git#1.0.1"
   }
 }
--- a/electron/package.json
+++ b/electron/package.json
@@ -21,10 +21,10 @@
     "open-graph": "0.2.1",
     "raygun": "0.9.0",
     "request": "2.76.0",
-    "winston": "https://github.com/wireapp/winston.git#2.2.0-e"
+    "winston": "https://github.com/wireapp/winston.git#2.2.0-e",
+    "spellchecker": "3.3.1"
   },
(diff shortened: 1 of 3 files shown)


lipis:
Since it will work on Windows.. move that to `dependencies` as well.. https://github.com/wireapp/wire-desktop/blob/master/electron/package.json#L28

electron/package.json
```diff
@@ -21,10 +21,10 @@
     "open-graph": "0.2.1",
     "raygun": "0.9.0",
     "request": "2.76.0",
-    "winston": "https://github.com/wireapp/winston.git#2.2.0-e"
+    "winston": "https://github.com/wireapp/winston.git#2.2.0-e",
+    "spellchecker": "3.3.1"
```
lipis: alphabetically :)

ConorIA:
I think that electron-builder has recently fixed the rebuild command for yarn. I pushed a PR that disabled re-building for the specific arch before packaging with electron-builder, but that could probably be revisted now.

mythsunwind:
@ConorIA can you link the PR here? I could not find it on their github. Would be interesting to have a look.

Currently we had to downgrade to use `npm install` for building versions instead of yarn because of https://github.com/yarnpkg/yarn/issues/2072 and because we cannot re-build for specific architectures with yarn.

lipis:
Does the spellchecker work for you @mythsunwind on Windows?

ConorIA:
@mythsunwind, as far as I can tell, the commits related to this issue (https://github.com/electron-userland/electron-builder/issues/861) should make rebuild compatible with yarn, but I haven't been able to make it work on my local install.

ConorIA:
@mythsunwind, @lipis, rebuilding is working with the latest yarn (v0.18.0) release.

lipis:
Woho!! I'm not entirely sure how to test and what was the problem.. but we could check in depth tomorrow!

As for the spellchecker it should work on Linux.. was working on my end earlier this week!

mythsunwind:
Will close this PR. Replaced by #248

lipis:
Good stuff...!

PR #242, opened 2016-12-02
Flash icon in taskbar on new messages (Windows)

mythsunwind:


lipis:
👍 nice one :)

gregor:
@lipis @mythsunwind Very nice and we definitely have to improve in this area but why did we merge right away? This is a design/interaction decision. I doubt we want the icon to be flashing the whole time... How are other messaging apps handling this?

lipis:
@gregor you have a point here.. but it's not flashy all the time but only for a few seconds on the first message.. so it's quite nice. Did you try it?

gregor:
No, not yet. I will give it a try today and compare the interaction with other apps.

lipis:
Sure! Let us know and maybe show it to Yorgos.. :)

PR #250, opened 2016-12-06
Move spellchecker dependency to non-optional dependencies

mythsunwind:


lipis:
still have issues with Windows builds in general though.. when you do a clean first install.. do you have any errors?

herrmannplatz:
We should sort the build issues and out and move it out of the optionals

PR #253, opened 2016-12-06
Ensure that the screen we are starting the wrapper on exists (WEBAPP-3346)

gregor:


electron/js/lib/pointInRect.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+'use strict';
+
+
+module.exports = function(point, rectangle) {
+  return point[0] > rectangle.x &&
+    point[0] < rectangle.x + rectangle.width &&
+    point[1] > rectangle.y &&
+    point[1] < rectangle.y + rectangle.height;
```
lipis: Don't we want to include if it's on the edge? `>=` and `<=`

lipis:
I think we should care only about the top left corner if it's in the view.. if not then just move it to `x = 0` if `x < 0` and `y = 0` if `y < 0`.

Because at least you will be able to move it around and put it somewhere else.. if we care about the bottomRight corner.. even if it's the view and the height of the screen is smaller than the height of our window.. we will have a problem by putting the window in the centre.

electron/js/lib/pointInRect.js
gregor: Sure

gregor:
After this mornings discussion I checked the return values of http://electron.atom.io/docs/api/screen/#screengetdisplaymatchingrect. As feared the returned display is not the one the upper left app corner sits on but where the majority of the window area resides. A check of only the upper left is not sufficient therefore. We will merge like this and monitor whether we have to improve the resizing process down the road through CS and other customer feedback channels.

PR #317, opened 2017-01-14
Fix spelling checkbox state in system menu

maximbaz:
As of today, when spelling is not configured in init.json, the spell check is not working. 
However in this special case the checkbox "Check Spelling while typing" is enabled, implying that the spell check _should work_. But it doesn't.

The intention obviously was to make spell check enabled by default, but it never worked.

This pull request fixes the checkbox state, so that it isn't misleading anymore. This preserves the current behavior of the app.

If you want to enable spell check by default, it can be achieved with the code below. I didn't do it as it would be a breaking change, people will suddenly start seeing a spell check when it wasn't there before. 

If you still want me to do it, suggest me where to place this code better, and I will do update this pull request. Ideally it should happen some time before `systemMenu.createMenu()` is called, so that the first time the app is loaded the checkbox already has a correct state.

```
if (typeof init.restore('spelling') == 'undefined') {
    init.save('spelling', true);
}
```

### DIFF
--- a/electron/js/menu/context.js
+++ b/electron/js/menu/context.js
@@ -215,7 +215,7 @@ if (config.SPELL_SUPPORTED.indexOf(locale.getCurrent()) > -1) {
   const spellchecker = require('spellchecker');
   webFrame.setSpellCheckProvider(locale.getCurrent(), false, {
     spellCheck (text) {
-      if (!init.restore('spelling')) {
+      if (!init.restore('spelling', true)) {
         return true;
       }
       selection.isMisspelled = spellchecker.isMisspelled(text);


lipis:
Good catch! What if we just add the default value `true` here as well instead of the opposite?
https://github.com/wireapp/wire-desktop/blob/master/electron/js/menu/context.js#L218

maximbaz:
It is also a possibility, but I'm not particularly fond of dragging this knowledge that default value must be `true` in all places where this configuration is used — next time someone will need to read this configuration, they are likely to again forget to use `true` as a default :)

Up to you though :)

lipis:
Every `init.restore()` call has a default value.. so as a compromise maybe we can have something like `config.SPELLING_DEFAULT` which will be set to `true` and we could use that instead of a magic value :) WDYT?

maximbaz:
Actually you are right and I didn't notice this before, every single call to `init.restore` in the codebase does have two parameters passed...

Then I agree, since it's almost like a rule and normally people will know to pass a second parameter, it's simpler to just pass `true`.

Given all other examples, making a constant for this one case only is probably an overkill :)

lipis:
Great! So let's just go with `true` and we are set :)

lipis:
🍻

PR #334, opened 2017-01-19
Fix init.restore() loading defaultValue when real value is 'false'

maximbaz:
A bug fix for #317, since the default value `true` is also used when the real config value is `false`. 

_In theory_ there is no way to disable spell check today.  In practice, the spell check anyway doesn't work in production on Linux (only?) :disappointed: due to this bug: https://github.com/atom/node-spellchecker/issues/49

/cc @lipis :)


ConorIA:
As a last resort, we could make the call to `electron-builder` disable asar. (See: https://github.com/atom/node-spellchecker/issues/49#issuecomment-254964968)

maximbaz:
Good point. Let's continue this topic in #335.

lipis:
I'll have to test those guys.. on Monday :) Thank you for the follow ups

gregor:
Totally makes sense - good catch

Issue #147, opened 2016-11-03
Suggestion: Disable ASAR packaging on Linux


ConorIA:
In #78, a Linux user asked to provide an extra tray icon for better integration on their Linux Mint desktop. I was poking around at the Electron-Builder options, and it seems that there is an option to disable ASAR  archiving when building. This would allow a savvy user to replace the tray.png and tray-badge.png images with something better integrated with their desktop. 

According to the Electron documentation, there are three main advantages to using ASAR packaging: 
1) Overcome long path names on Windows (not as much of an issue on Linux)
2) Slightly increase the speed of require commands (might be relevant)
3) Hide source code (not necessary, as Wire is already all out in the open on GitHub!)

ConorIA:
I take it from the lack of response from either other users or from the devs that this isn't something anyone really wants or needs. I know I, for one, would probably not even use it. I will close this now, but should anyone happen upon this issue later and feel strongly about it, feel free to ping me and I'll re-open it.

PR #335, opened 2017-01-19
Disable asar for Linux

maximbaz:
This is a temporary workaround for atom/node-spellchecker#49, as suggested by @ConorIA in #334. 

I confirm that it does fix the issue.

Disabling only for Linux since the original bug describes this as a Linux-only issue, though I didn't verify this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "linux" : {
       "afterInstall" : "bin/deb/after-install.tpl",
       "afterRemove" : "bin/deb/after-remove.tpl",
+      "asar": false,
       "category": "Network",
       "depends": ["libappindicator1", "libasound2", "libgconf-2-4", "libnotify-bin", "libnss3", "libxss1"],
       "target": ["AppImage", "deb"]


ConorIA:
So fast! See also: #147.

ConorIA:
Strangely on my machine I still can't get the spellchecker to work. I wonder what I'm missing.

ConorIA:
I figured out what was wrong on my machine ... no love for Canadian English! But, when we disable asar, I can use the same solution as the packager for Atom on Arch used (https://github.com/atom/spell-check/issues/129#issuecomment-223385570).

ConorIA:
Note that instead of `"asar": false`, we could use `"asarUnpack": "node_modules/spellchecker"`. 
In fact, `electron-builder`'s wiki says that node modules that need to be unpacked will be automatically detected, so I will report this as an issue over there.

lipis:
So what do you guys say.. shall I merge this one? Will it work?

lipis:
Spell checker works on my Ubuntu when running it with `npm start` but I can't build it :/

maximbaz:
This definitely fixes the issue for me for the release build.

> Note that instead of "asar": false, I used "asarUnpack": "node_modules/spellchecker/vendor".

I'd like to hear @ConorIA's comments on this, should we go with the `asarUnpack` instead? Does it have some benefits? What would you prefer?

And in general, I'm not sure if we are treating this as a temporary workaround, or we will leave it permanently as asked in #147? If we keep it forever for the sake of #147, then the `"asar": false` is the way to go, I think.

lipis:
Updated everything and it looks like it worked for me as well. So I think I'll go ahead and merge this one.

lipis:
If anything we can fix it later! Thanks guys..

ConorIA:
@maximbaz, I just offered `asarUnpack` as an alternative because `electron-builder` makes noise when we build with asar disabled: 

```
⚠️  Packaging using asar archive is disabled — it is strongly not recommended.
Please enable asar and use asarUnpack to unpack files that must be externally available.
```

I think a full disable will give us the opportunity to appease tray-icon sticklers though, so I think the current merge is a safe bet.

PR #336, opened 2017-01-20
Improve auto-updater behavior on Windows (WEBAPP-3544)

gregor:
*This targets the auto update behavior on Windows only*

Currently we only check for a new version of our Windows app on app start and silently install it in the background for use on next app start.

- we will now check for app updates on app start and subsequently every 12 hours
- once a version is installed an bar informing the user will be shown
- a click on said bar restarts the wrapper


lipis:
Cool stuff!

electron/js/squirrel.js
```diff
@@ -162,11 +161,34 @@ function removeStartupShortcut(callback) {
...
+    if (update) {
+      installUpdate();
+    }
+  });
 }
 
 
+function installUpdate() {
+  spawnUpdate(['--update', config.UPDATE_WIN_URL], function(error) {
+    if (error != null) {
```
lipis: `!==`

lipis: or actually with `null` I think it should be as is.. nevermind..

PR #337, opened 2017-01-21
Preferably retrieve BrowserWindow by ID

gregor:
I have seen electron actions performed in the BrowserWindow fails because the currently we rely on the focussed window. Sometimes focus is on the about screen or devtools and actions thus fail resulting in errors.

This should be safe in the menu cases but I temporarly reused the code when making changes to the updater and came across this problem. This can be handled in a more robust way. I am thus disclosing the window ID of the main BrowserWindow and retrieve it based on the ID instead of focus.

### DIFF
--- a/electron/js/window-manager.js
+++ b/electron/js/window-manager.js
@@ -1,0 +1,40 @@
+/*
+ * Wire
+ * Copyright (C) 2016 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


electron/js/window-manager.js
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+'use strict';
+
+const {BrowserWindow} = require('electron');
+
+let primaryWindowId;
+
```
lipis: maybe two lines before/after functions

(I just noticed that we are not that consistent in this project)

electron/js/window-manager.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+
+function getPrimaryWindow() {
+  if (primaryWindowId) {
+    return BrowserWindow.fromId(primaryWindowId);
+  }
+};
+
+module.exports = {
```
lipis: here as well maybe.. 2 empty lines.

PR #350, opened 2017-01-24
Make fingeprints selectable

### LINKED ISSUE #320: Can't copy device fingerprint
When I go to my devices I can't select the fingerprint to copy/paste it.

This is on Linux.

lipis:
Fixes #320


herrmannplatz:
We should introduce a class to explicitly make things copyable. You could even fix that in the webapp

lipis:
Something like this: https://github.com/wireapp/wire-webapp/pull/681?

Issue #355, opened 2017-01-24
No notifications


lightswitchr:
I cannot get notifications on Windows to work at all, on any notification setting. The sound does play.

Windows 7, Wire version 2.11.2673

GopianiS:
Same for me talking to lightswitchr and dont receive popup notifications

gregor:
There is no notifications support in Windows 7.

lightswitchr:
@gregor - Why is that? You have a notification icon in the tray, it is trivial to provide balloon notifications through that.

lightswitchr:
See [this in the Electron documentation](http://electron.atom.io/docs/api/tray/#traydisplayballoonoptions-windows).

gregor:
We have a limited amount of resources and implementing a feature like this for legacy OS support was never prioritized high enough to actually implement it. It is currently not on our list. I am sure we would accept a contribution in this area though and are always evaluating priorities taking user demand into account.

lipis:
Windows 7 is already a 7 years old OS.. what stops you to upgrade to something newer?

lightswitchr:
@gregor, I am aiming to make a pull request, however npm install is failing for me at the moment.

@lipis - work licences. Also, I like it, and according to [this source](https://netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0) Windows 7 still has a 48% market share. I would hardly call it legacy.

lipis:
There are at least 3 major version after Windows 7.. and Microsoft is trying hard to get all those people upgrade to something newer!

lightswitchr:
All I want is notifications from a chat app, shouldn't need a whole new operating system for that :)

lipis:
Point taken :) and yes.. we are off topic.. :)

lightswitchr:
@gregor @lipis I have the balloon working - I've been trying to push the branch but I get this:

> ERROR: Permission to wireapp/wire-desktop.git denied to lightswitchr.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.

How do you want me to proceed?

lightswitchr:
Never mind, I found out how. Pull request created.

lipis:
As we are not able to spend resources on this one, the best way is to follow this issue on Electron's issue tracker: https://github.com/electron/electron/issues/8384. I hope it will be OK to close it here.

PR #357, opened 2017-01-24
Adding balloon notifications for Windows #355

lightswitchr:
This adds balloon notifications to Windows. It may have to be narrowed down so that it only fires on Windows 7.

### DIFF
--- a/electron/js/menu/tray.js
+++ b/electron/js/menu/tray.js
@@ -67,9 +67,9 @@ module.exports = {
       if (count) {
         this.useBadgeIcon();
         if (count > lastUnreadCount) {
-            this.displayBalloonNotification(count);
-            }
-        } else {
+          this.displayBalloonNotification(count);
+          }
+      } else {
         this.useDefaultIcon();
       }
       win.setOverlayIcon(count && process.platform === 'win32' ? iconOverlayPath : null, locale.getText('unreadMessages'));
@@ -90,9 +90,9 @@ module.exports = {
   },
   
   displayBalloonNotification: function(count) {
-      if (appIcon == null) return;
-      var options = {title:'Wire', content:'You have unread messages in ' + count + ' conversation(s).'};
-      appIcon.displayBalloon(options);
+    if (appIcon == null) return;
+    var options = {title:'Wire', content:'You have unread messages in ' + count + ' conversation(s).'};
+    appIcon.displayBalloon(options);
   },
   
 };
--- a/electron/js/menu/tray.js
+++ b/electron/js/menu/tray.js
@@ -68,7 +68,7 @@ module.exports = {
         this.useBadgeIcon();
         if (count > lastUnreadCount) {
           this.displayBalloonNotification(count);
-          }
+        }
       } else {
         this.useDefaultIcon();
       }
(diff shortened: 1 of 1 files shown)


maximbaz:
Just for my curiosity, how will it behave on systems which do show notifications today? Will two notifications appear now, a native one and this balloon? I kind of expected to see some `if-else` in the code, like "if native notifications are not supported, use balloon", but maybe I am simply missing something?

I tried on my ArchLinux, I don't have duplicate notification (which is good), but that may be because such balloons are not supported for Linux at all. If anyone could try on Windows 10 for example, that would be interesting.

gregor:
@lightswitchr Thank you very much for opening this PR and your quick contribution in response to my comment on the issue #355. 

When evaluating contributions from amazing loyal users like you, we have to keep the needs and expectations of all our users in mind. In the end we want to deliver a great experience to as many users as possible given our resources. The majority of Windows users has at this point upgraded to platforms released after Windows 7 hit the market many years ago. And while Windows 7 retains a large market share in business environments, our user base shows a strong tendency to newer platforms. We are focused on delivering a rich, comprehensive end-to-end experience on those platforms. If we add features targeted at legacy platforms (mainly in the form of user contributions as our current priorities are different) we have to ensure that the user experience does not  deteriorate on other platforms. While I can fully understand your need for a richer experience on Windows 7 - as it is the platform you are using - I ask you take a step back for an take a broader view of how other user would be affected by the proposed change. Adding features usually turns out to be a lot more complex than it might seem on first look.

After having evaluated the impact of this PR only from a very broad perspective, we already see three major concerns with the proposed approach:

1. We would end up with balloon notifications on all Windows platforms not matter whether they do support and make use of more advanced notifications - just like @maximbaz pointed out.

2. The current implementation ignores user preferences, which are handled in our [wire-webapp](https://github.com/wireapp/wire-webapp) project. The version released to [app.wire.com](https://app.wire.com) is loaded inside the [wire-desktop](https://github.com/wireapp/wire-desktop) apps at runtime. The webapp preferences must be followed. In the proposed approach users would get balloon notifications even if they selected to never show any notifications. The set of available preferences concerning notifications currently available is quite comprehensive and should be reflected in the behavior of the feature when adding new styles of notifications.

3. In the current approach the balloons are generic and do not actually contain any notification content. The information portrayed by them is handled by other user interaction models in Windows builds. The same information is currently available in Windows 7 with the overlay in the system tray icon. We have updated the style of this icon with a more prominent red dot to be more obvious, the taskbar icon is also getting an overlay and will flash on new messages in additional chats. These should increase the visibility of unread messages even if the systray icon is not visible as the bar collapsed. Due to issues with electron we have not released a public build since merging these changes in early December. But we are getting really close.

Adding balloon notifications with detailed message content adhering to the webapp preferences and only show on Windows 7 is definitely possible. I hope I was able to point out the complexity of this feature though and what is required to add it to Wire in order to maintain the comprehensive user experience we always strieve for - and why it would require more resources than we are able and willing to commit to build this improvement ourselves at this point. 

To conclude, we cannot merge your contribution in its current form to the project. If you want to contribute and are willing to invest more time in an effort to evolve the contribution, I am more than happy to help you out where I can. It definitely won't be an easy thing to add though as it might even need changes on both [wire-webapp](https://github.com/wireapp/wire-webapp) and [wire-desktop](https://github.com/wireapp/wire-desktop).

lightswitchr:
Well it was worth a try. If you've already got other things in the pipe line then I won't bother. I will keep the fork and use it myself for the time being until you release a new public build.

PR #362, opened 2017-01-25
Remove all but the last line from RELEASES file

gregor:


lipis:
The shorter I could come up with is

```python
with open('RELEASES', 'r+') as f:
  f.readline()
  f.truncate()
```

gregor:
@lipis Just tested this. We need the last line of the file and not the first one though.

gregor:
Code is more like that. Another puzlle to shorten?

````python
with open('RELEASES', 'r+') as f:
  last = f.readlines()[-1]
  f.seek(0)
  f.write(last)
  f.truncate()
````

lipis:
if it's last.. then it looks like that's the one.. :)

PR #368, opened 2017-01-26
Improve Windows update experience

gregor:


gregor:
Need to test this in an internal build. Will merge now, release and check while internal release for Windows still needs to be changed manually.

PR #390, opened 2017-02-01
Backwards compatibility for win32 startup

lipis:


electron/js/menu/system.js
```diff
@@ -443,6 +444,13 @@ module.exports = {
     }
 
     if (process.platform === 'win32') {
+      if (squirrel.checkForOldStartup()) {
```
gregor: Do we need to check this on every app start or can we rely on the squirrel handler update case instead?

PR #391, opened 2017-02-01
".eslintrc" is deprecated

bennyn:
http://eslint.org/docs/user-guide/configuring#configuration-file-formats


lipis:
I would prefer `.yaml` though :)

http://www.yaml.org/faq.html and also from the above link the first recommendation is `.yaml`.. even if that means that we have to change `travis.yml` to `travis.yaml` :D

PR #414, opened 2017-02-07
expose winston in process load

herrmannplatz:


electron/js/preload.js
```diff
@@ -19,8 +19,10 @@
 
 'use strict';
 
-const {ipcRenderer, webFrame, desktopCapturer} = require('electron');
+const {remote, ipcRenderer, webFrame, desktopCapturer} = require('electron');
+const {app} = remote;
```
bennyn: `const app = remote;`?

lipis: or consistency :)

herrmannplatz: ?

PR #417, opened 2017-02-07
Logging the log file path

bennyn:
Within the Chrome DevTools I want to see the log file path to which the logs are written.


electron/js/preload.js
```diff
@@ -35,9 +35,12 @@ process.once('loaded', function() {
   global.desktopCapturer = desktopCapturer;
   global.winston = require('winston');
 
+  const logFilePath = path.join(app.getPath('userData'), require('./config').CONSOLE_LOG);
+  console.log('Logging into file: ' + logFilePath);
```
lipis:
```
...('Logging into file', logFilePath);
```

lipis: but do we really want that?

Issue #416, opened 2017-02-07
Indicator icon notification


fungs:
I use the wire app under Linux and it is really easy to miss post or calls when you are not looking at the screen or coming back to your computer after a while. So IMO wire should be able to (a) change the tray indicator icon if new calls or messages arrived and (b) be able to pop up the application when minimized to the tray.

ConorIA:
The tray icon does change, but the current published build contains a bug preventing it from changing. It is fixed in the upstream code.

fungs:
Then, I suppose to add the popup feature as well, that is how many chat programs behave (making it optional). Thanks for the great app!

maximbaz:
But the desktop notifications are also supported, unless you turned them off in the settings.

Besides the tray icon and desktop notifications, Wire also sends urgency hint (at least, for now #371), which if interpreted by your desktop environment provides yet another way to notify you about updates.

fungs:
I get your point @maximbaz, the desktop should handle the notification, fine. However, it's quite inconvenient that I have to double-click the tray icon to pop up the chat window, which in my case opens minimized, so I have to double-click, then click on the minimized window to open it. This might seem minor, but if you have to do this really often per day, this is annoying.

Using MATE desktop environment.

maximbaz:
Sorry I guess I simply don't understand what you mean by "popup feature" then :)

You want wire to be automatically maximized and brought in the foreground as soon as any message appears?

fungs:
> You want wire to be automatically maximized and brought in the foreground as soon as any message appears?

This is how many messengers work, because you don't need to switch from keyboard to mouse to keyboard.

However, the minimum would be that the message windows opens **not minimized** when you (double)click on the tray icon, so that you are ready to type.

Please note: I'm not talking about maximizing the window, it should just not be minimized and it should be the active window.

maximbaz:
> This is how many messengers work...

Interesting, thanks for explanation - I really didn't know that. I assume these messengers usually have a pretty narrow window? It's just hard to imagine that it's not annoying when you are browsing internet and every 5 seconds your browser is suddenly becoming fully covered by a messenger window. But if it's a small narrow window on the side, I can see how this can be useful...

> However, the minimum would be that the message windows opens not minimized when you (double)click on the tray icon, so that you are ready to type.

Yeah, now I get what you mean, that definitely sounds like an issue to be fixed.

fungs:
[...]

PR #419, opened 2017-02-07
Restore minimized window on tray click (#416)

ConorIA:
This is intended to resolve some of the frustration expressed in #416. I have done some local testing, and it seems to be working as expected. ~~Of course, Gnome has some strong focus-stealing protection, so even when I click on the tray badge, it gives me a "Wire is ready" notification instead of bringing the window into focus.~~ It should probably be tested on other DEs.

Update: I was browsing some other [FLOSS code](https://github.com/nylas/nylas-mail/blob/3111c166b8c949cf2a5ef60392d83a53e782ebed/src/browser/window-manager.es6) (shoutouts to the Nylas folks!), and copied their solution. This should work now. 

~~Should we stick the `windowManager.getPrimaryWindow()` into a `var` though? (or `const` or `let` ... still don't know what the role of each is) The way it is written now, we're calling the function no less than 8 times!~~

Update 2: I tried sticking this into a function in the `windowManager.js` file. ~~I am not sure about passing arguments in Javascript. I imagine there is some way to call `showPrimaryWindow` on click without wrapping it in an additional function to use `windowManager.getPrimaryWindow()`. I tried a couple of things, e.g. putting the call to `getPrimaryWindow()` inside the `showPrimaryWindow` function, but it gave me errors.~~

Update 3: I guess I can't use a named function as the `click:` response? All seems to be in order (more or less) now. Happy to get feedback and test results.


gregor:
@lipis If this works we should also hook it up to the notification clicks in https://github.com/wireapp/wire-desktop/blob/5d759b0e17340ead54e9a3a543c2bbf351a23b38/electron/main.js#L127 as people have been complaining that the app is not always put to the foreground then.

ConorIA:
@gregor, I've just added an additional commit that sticks this into two spots in `main.js`: on notification click, and on launching a second instance.

gregor:
@ConorIA Thanks for adding it to the mentioned use cases as well. Back from vacation and reliving what I missed while I was gone just now.

ConorIA:
No problem. Hope you had a relaxing vacation! 🌊🚢🌴

Issue #397, opened 2017-02-02
No portable version?


ghost:
Only installer version is a show stopper.

related: qtox.github.io provides installer AND portable.

lightswitchr:
Don't need a portable version. Use https://app.wire.com instead.

maximbaz:
For curious minds, there is a way to package app in [AppImage](http://appimage.org/) format, which is basically a portable version for Linux. 

That being said, I double the suggestion to just use website.

ConorIA:
@maximbaz, I almost suggested the AppImage too (btw Wire provides AppImage bins on the download page!) but figured the request was more about Windows. 

In terms of privacy, does the webapp cache anything on the host system that wouldn't get flushed with a simple "clear private data"?

maximbaz:
> btw Wire provides AppImage bins on the download page!

Wow I've never noticed that, cool!

> In terms of privacy, does the webapp cache anything on the host system that wouldn't get flushed with a simple "clear private data"?

You will find this page interesting to read: https://github.com/wireapp/wire-webapp/issues/688
In short, if you don't enable "Remember me", nothing will be cached after "Log out".

githubbantor:
webpage version is not usable via tbb. https://github.com/wireapp/wire-webapp/issues/722

githubbantor:
I'm talking about zip/rar/7z format, not AppImage. I'm expecting a windows portable version.

sergeevabc:
+1 for Windows Portable.

helenakos:
Thanks for your request, we don't have it planned at the moment so closing this issue. 

On a side note, you can open an issue for general comments (not related to any client / specific repository) here: https://github.com/wireapp/wire

PR #427, opened 2017-02-12
Add portable flag (#397)

ConorIA:
This is a quickfix for #397. Adds a "portable" command line flag that moves the `userData` dir to `APP_PATH/../WireData`. I put it in the parent of `APP_PATH` so that we don't risk wiping the folder on re-install or update. Not sure how this will work with the Squirrel stuff, but might at least be a nudge in the right direction. Ideally, the user could install the app onto a usb key under a "WirePortable" folder, which would then contain the installed folder and the WireData folder.

Proof that it works: 
```
$ git status
On branch portable
Your branch is up-to-date with 'origin/portable'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	WireData/

nothing added to commit but untracked files present (use "git add" to track)
```

Also, running with the `--portable` flag asks me to log in, which means that Wire is not reading from my usual `~/.config/Wire` directory.


ConorIA:
It didn't occur to me to try this in an actual install. Since app dir is `<basedir>/resources/app`, the `WireData` dir is getting stuck into `<basedir>/resources/WireData`. Do we like that? Should it be in the basedir or perhaps even in the directory above the basedir? If Wire is updated, does the `resources` dir get wiped out? Or are the files just updated in place?

lipis:
Since it's suppose to be portable.. I would add it to the same directory with the app (where the executable resides). WDYT?

ConorIA:
There it is! It now resides beside `process.execPath`. It just occured to me that this isn't going to work for AppImage, though admittedly, I couldn't get the AppImage to accept the `--portable` flag even before choosing to use `process.execPath`.

electron/main.js
```diff
@@ -68,6 +68,12 @@ if (config.DEVELOPMENT) {
   app.commandLine.appendSwitch('ignore-certificate-errors', 'true');
 }
 
+if (argv.portable) {
+  const EXEC_PATH = (process.env.APPIMAGE != null) ? process.env.APPIMAGE : process.execPath;
+  const USER_PATH = path.join(EXEC_PATH, '..', 'WireData');
```
lipis: This will generate the `WireData` directory next to `Wire`.. right?!

I was thinking.. why not having something like that: 

```js
const USER_PATH = path.join(EXEC_PATH, 'Data');
```

electron/main.js
```diff
@@ -68,6 +68,12 @@ if (config.DEVELOPMENT) {
   app.commandLine.appendSwitch('ignore-certificate-errors', 'true');
 }
 
+if (argv.portable) {
+  const EXEC_PATH = (process.env.APPIMAGE != null) ? process.env.APPIMAGE : process.execPath;
```
lipis:
```js
const EXEC_PATH = process.env.APPIMAGE || process.execPath;
```

ConorIA:
Yep. That puts the WireData folder beside the executable. I think the '..' is necessary, because we want the data in the same folder as the executable. I haven't tried, but I think without that the app would try to create the Data folder under the executable, not beside it. I'm on the bus now, but can change WireData to Data when I get home.

lipis:
Yes.. I think just `Data` is enough.. if the folder is already inside the `Wire` directory.

lipis:
@ConorIA is it working as expected on Linux? We'll test it with our Windows folks tomorrow and we could merge it.

ConorIA:
I only tested a wee bit, but it seemed to work as I had hoped.

ConorIA:
Still works! 😉

PR #371, opened 2017-01-27
Don't bounce on macOS on new messages

lipis:

PR #448, opened 2017-02-19
Add macOS tray (menubar) icon support

sidneys:
Imgur

Dear Wire team,

I added image assets and necessary logic for the macOS tray/menubar icon, as the "bouncing" behaviour for incoming messages was recently removed, leaving no visual cues for unread messages.

The implementation is based on [Electron's](https://github.com/electron/electron/blob/master/docs/api/native-image.md) [Template-Image](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/ToolbarIcons.html)-aware asset handling logic.

Please review and share your thoughts and communicate change requests as required.
Thanks for a great messaging platform.

Cheers,
S

P.S.: The existing tray menu structure does not follow [the macOS platforms' UI best practices](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/MenuBarMenus.html). This should be improved at a later point.


lipis:
The bouncing behaviour is not removed, this issue is closed without merging: https://github.com/wireapp/wire-desktop/pull/371.

But the bouncing behaviour didn't make it to the Apple Store yet.. so that's why you don't see it. As for the menu icon, it was deliberately not included (as it is on other platforms) because it's already visible in many places.

sidneys:
@lipis Thanks for your input!

However, the inclusion of a menubar icon, especially on macOS, and can be considered a common feature for communication-focus apps. It's implemented by Apple in their own Messages.app application,  as well as many wildly successful third-party messengers, e.g. for WhatsApp ([App for WhatsApp](https://itunes.apple.com/de/app/app-for-whatsapp/id963574990?l=en&mt=12)), Google Hangouts ([YakYak](https://github.com/yakyak/yakyak)) or Facebook Messenger ([Messenger for Desktop](https://github.com/Aluxian/Messenger-for-Desktop)).

Furthermore, a bouncing dock icon does not follow macOS UI guidelines, as [it refers to a warning or error that must be attended to immediately, rather than a simple notification](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/Dock.html). This would require a badge icon on the Dock, rather than a bounce.

As an aside, the missing Wire menu icon on the Mac has often sparked questions from non tech-savvy colleagues of mine, who prefer a less cluttered and distracting Dock area. Users are less likely to close the app if it runs in the background, thus encouraging continuous use.

Please consider including this feature, as with the other platforms.

Cheers,
S

helenakos:
@sidneys Hi, thanks for the request. We considered it (once again) and discussed internally. However we think that menu bar icon is for utility apps and jobs you do often. Having your unread messages notification in Dock is something we will stick to. 

You can read more in [macOS guidelines](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/MenuBarMenus.html) on the topic: "Consider alternatives to creating a menu bar extra. For example, you can use the Dock menu functions to open a menu from your app’s icon in the Dock."

Sytten:
@sidneys hey there, I was asking the same thing haha
Its the first thing I notice when i opened the app: no menu bar icon, wtf.
I don't see why they don't want to add it (or at least give the option in the setting).
It's such a common feature on mac...

PR #460, opened 2017-02-22
Bump electron version

herrmannplatz:
electron recently changed the the `setCertificateVerifyProc`api. so in order to update to 1.6.1 we need update the certificate verification code on our side.

see: https://github.com/electron/electron/blob/master/docs/api/session.md#sessetcertificateverifyprocproc

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -179,7 +179,7 @@ function showMainWindow() {
     if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, cert))) {
       cb(-2);
     } else {
-      cb(0);
+      cb(-3);
     }
   });
 
--- a/electron/main.js
+++ b/electron/main.js
@@ -172,7 +172,7 @@ function showMainWindow() {
   main.webContents.session.setCertificateVerifyProc(({hostname = '', certificate = {}}, cb) => {
     const {data: cert = ''} = certificate;
 
-    if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, certificate.data))) {
+    if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, cert))) {
       cb(-2);
     } else {
       cb(0);
--- a/electron/main.js
+++ b/electron/main.js
@@ -169,13 +169,13 @@ function showMainWindow() {
     main.setBounds(init.restore('bounds', main.getBounds()));
   }
 
-  main.webContents.session.setCertificateVerifyProc((hostname = '', certificate = {}, cb) => {
+  main.webContents.session.setCertificateVerifyProc(({hostname = '', certificate = {}}, cb) => {
     const {data: cert = ''} = certificate;
 
-    if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, cert))) {
-      cb(false);
+    if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, certificate.data))) {
+      cb(-2);
     } else {
-      cb(true);
+      cb(0);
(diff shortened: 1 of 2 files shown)


electron/main.js
```diff
@@ -169,13 +169,13 @@ function showMainWindow() {
...
-  main.webContents.session.setCertificateVerifyProc((hostname = '', certificate = {}, cb) => {
+  main.webContents.session.setCertificateVerifyProc(({hostname = '', certificate = {}}, cb) => {
     const {data: cert = ''} = certificate;
 
     if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, cert))) {
-      cb(false);
+      cb(-2);
     } else {
-      cb(true);
+      cb(0);
```
lipis: Shall we maybe use some more meaningful names :)

```js
config.CERTIFICATE_FAILED_CATASTROPHICALLY = -2
config.CERTIFICATE_OK = 0
```

lipis: even if it's not inside the `config`

electron/main.js
```diff
@@ -169,13 +169,13 @@ function showMainWindow() {
     main.setBounds(init.restore('bounds', main.getBounds()));
   }
 
-  main.webContents.session.setCertificateVerifyProc((hostname = '', certificate = {}, cb) => {
+  main.webContents.session.setCertificateVerifyProc(({hostname = '', certificate = {}}, cb) => {
     const {data: cert = ''} = certificate;
```
ffflorian: Now that `request` is an object, we can use 
```js
const { hostname = '', certificate: { data: cert = '' } = {}, error } = request;
```
so we don't destructure twice. I think it's more consistent.

And check for error maybe?

herrmannplatz: I guess in case of an error we should invoke the the callback with 0.

Need to get used to this nested destructering :)

herrmannplatz:
Seems like we can simplify the certificate validation. But i would do this in a different PR. This should be good to go.

electron/main.js
```diff
@@ -169,13 +169,17 @@ function showMainWindow() {
     main.setBounds(init.restore('bounds', main.getBounds()));
   }
 
-  main.webContents.session.setCertificateVerifyProc((hostname = '', certificate = {}, cb) => {
-    const {data: cert = ''} = certificate;
+  main.webContents.session.setCertificateVerifyProc((request, cb) => {
+    const { hostname = '', certificate: { data: cert = '' } = {}, error } = request;
+
+    if (error !== undefined) {
+      return cb(0);
```
ffflorian:
```js
return cb(-2);
```

electron/main.js
ffflorian: No, better invoke with `-2`, otherwise it will be okay if the certificate couldn't be checked. :)

electron/main.js
raphaelrobert: If pinning checks out, we should probably do

`return cb(-3);`

I think it matches what the old `return cb(true)` did.

ffflorian: Good idea - in this case then we don't need to double-check because the certificate won't be valid anyway.

herrmannplatz: yeah my fault 0 is success

electron/main.js
```diff
@@ -169,13 +169,17 @@ function showMainWindow() {
     main.setBounds(init.restore('bounds', main.getBounds()));
   }
 
-  main.webContents.session.setCertificateVerifyProc((hostname = '', certificate = {}, cb) => {
-    const {data: cert = ''} = certificate;
+  main.webContents.session.setCertificateVerifyProc((request, cb) => {
+    const { hostname = '', certificate: { data: cert = '' } = {}, error } = request;
```
lipis: La SpaCeS :)

Issue #442, opened 2017-02-16
Linux: Debian (9.0) wire package name: please change! (wire-desktop)


keypey:
Hi.

Debian 9.0 has "wire" package (see: https://packages.debian.org/stretch/wire).

Please, change wire package name (wire-desktop?).

--
kp.,

Strit:
That does not sound like this Wire. It sounds to be a library of coding protocols.

keypey:
It is true! But Wire ("wire-desktop") [wire_2.11.2685_amd64.deb] has this same name "wire" too!
So: please change [Wire ("wire-desktop") [wire_2.11.2685_amd64.deb]] name to "wire-desktop".

--
kp.,

ffflorian:
I actually second that - it would be just a simple change in the [`package.json`](https://github.com/wireapp/wire-desktop/blob/master/electron/package.json#L2). In the current state it is [quite hard](http://askubuntu.com/questions/326288/install-two-applications-with-the-same-name) to install the "Simple Wire encoding protocol command line tool" on Debian while Wire Desktop is installed (and vice versa).

What do you think @bennyn?

lipis:
What should be the name though.. I think that will be the toughest one.. :)

lipis:
### wire-app

lipis:
### wire-desktop

bennyn:
In terms of consistency `wire-desktop` wins (if you ask me) because it's written all over the place (and it's the GitHub repository name). Looking at the `package.json`, our current value for `name` (which is `wire`) would indicate that our project is a light, fast and flexible JavaScript IoC container (https://www.npmjs.com/package/wire).

ConorIA:
@lipis even with #463, the package is still building for me as `wire_2.11.2721_amd64.deb`. Not sure why, but it seems at some point in the build process the name in electron/package.json goes back to `wire`. Is this just happening on my machine?

lipis:
right.. but is that an actual problem?

ConorIA:
Well, it is insofar as the changes you made do not produce a Debian package called "wire-desktop" unless the name is also changed in `info.json`.

ffflorian:
@lipis 

The problem is not the filename - the package `wire` already [exists in a Debian repository](https://packages.debian.org/stretch/wire):
```
 $ dpkg --info wire_2.11.2685_amd64.deb | grep Package
 Package: wire
[...]

Issue #430, opened 2017-02-13
PPA for stable builds


darkwyrm:
Could it be possible to publish Linux .deb files via a PPA so that users can get updates using the package manager?

ConorIA:
Once set-up, this might be as simple as adding an extra task to the build process.
https://help.launchpad.net/Packaging/PPA/Uploading

anatoli26:
Any ETA for this?

RebelCoderRU:
We absolutely need PPA/Repo for .deb packages. for example, I installed Wire .deb file on 3 computers of my relatives + 2 computers of non-technical friends.

Now, to update to a latest build, I have to remote-in to 5 PCs, download new .deb file and install it. this leaves bad negative taste in generic users mouth.

Please organize this ASAP, so we can just install the first .deb file and it will link to your repos automatically, like 99.9% of other apps do.

anatoli26:
@lipis, @gregor, could this be setup in the nearest future?

lipis:
This definitely something that we are looking into it.. there is no committed date yet but we would like ta have it.

aquatooth:
Yes please! Definitely look into setting up a PPA to publish Linux .deb files so that users can get updates using the package manager.
Users of Skype that are or have migrated to Wire really do need this as soon as possible.

RebelCoderRU:
Agree! To much hassle for "moms and paps" kind of users without updates.

aquatooth:
Manually updating a number of installations with version 2.12.2729 today

lipis:
Point taken.. :)

aquatooth:
ah, damn good to know your team is on this issue, as it will cause trouble for every single install on Linux.
in this case, i must remote into several boxes in order to upgrade it, but after PPA it will be 'set and forget' So, thank you and your team!

RebelCoderRU:
Huston... Huston... we are loosing potential users.......

RebelCoderRU:
We still need PPA/Repo :/

aquatooth:
@RebelCoderRU Yes, posting in agreement especially as @ConorIA has posted what seems a straightforward method of ending the Linux manual update misery :)

RebelCoderRU:
@aquatooth Sorry, but I am not sure If I understood what you meant... Is this being done now ?

aquatooth:
@RebelCoderRU agree that your comment; 'We still need PPA/Repo'
[...]

PR #464, opened 2017-02-24
Change package name to avoid conflicts

lipis:
Round two on #442

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -192,7 +192,7 @@ module.exports = (grunt) ->
     electron_pkg = grunt.file.readJSON ELECTRON_PACKAGE_JSON
     electron_pkg.updateWinUrl = info.updateWinUrlInternal
     electron_pkg.environment = 'internal'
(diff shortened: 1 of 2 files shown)


ffflorian:
When installing the app as `wire-desktop` while having `wire` installed, I get this error:
```
Unpacking wire-desktop (2.11.2721) ...
dpkg: error processing archive wire-desktop_2.11.2721_amd64.deb (--install):
 trying to overwrite '/opt/Wire/libffmpeg.so', which is also in package wire 2.11.2721
dpkg-deb: error: subprocess paste was killed by signal (Broken pipe)
Errors were encountered while processing:
 wire-desktop_2.11.2721_amd64.deb
```

This means I have to uninstall `wire` first and thus losing all local data. As @lipis reported, it also changes the filename on Windows...

lipis:
We might want to put a hold on this one.. :(

lipis:
As it's it will delete the history.. we might deal with it when fixing the PPA package #430.

mythsunwind:
PR was replaced by #550

PR #506, opened 2017-03-10
Make the preference to hide the menu sticky

ConorIA:
A couple of questions here: 

1. Should the menu item be a check box?
2. Should the menu item be changed to "Hide the menu bar by default"?

As it stands now, a user might choose to hide them menu bar. If that user hits 'Alt', the user *might* think that the menu bar will now remain visible, however it will disappear again on restart, unless the user remembers to click on the "Show/Hide Menu Bar" item again. Maybe this isn't intuitive.

### DIFF
--- a/electron/js/menu/system.js
+++ b/electron/js/menu/system.js
@@ -223,6 +223,8 @@ let showWireTemplate = {
 
 let toggleMenuTemplate = {
   i18n: 'menuShowHide',
+  type: 'checkbox',
+  checked: !init.restore('hideMenu', false),
   click: function() {
     let mainBrowserWindow = getPrimaryWindow();
     if (mainBrowserWindow.isMenuBarAutoHide()) {
--- a/electron/js/menu/system.js
+++ b/electron/js/menu/system.js
@@ -227,9 +227,11 @@ let toggleMenuTemplate = {
     let mainBrowserWindow = getPrimaryWindow();
     if (mainBrowserWindow.isMenuBarAutoHide()) {
       mainBrowserWindow.setAutoHideMenuBar(false);
+      init.save('hideMenu', false);
     } else {
       mainBrowserWindow.setAutoHideMenuBar(true);
       mainBrowserWindow.setMenuBarVisibility(false);
+      init.save('hideMenu', true);
     }
   },
 };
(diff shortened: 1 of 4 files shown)


maximbaz:
1. Yes
2. Maybe just "Show menu", unchecked by default. It's self-explanatory, I think

Not related directly to your change, but I think the scenario of using <kbd>Alt</kbd> to show menu is a bit unusable now anyway, since menu accelerators are missing (those underlined characters on every top-level menu item, that can be used to open a particular menu). I hit <kbd>Alt</kbd> with a keyboard, but I can't do anything else with a keyboard, I have to grab a mouse.

image

lipis:
I would also go with the `Show Menu` but I'll ask our designers tomorrow for the final label. As for the default I would have it visible..

maximbaz:
Me too, that was a typo 😄

lipis:
@ConorIA 2. Change it to `Show Menu`

electron/js/menu/system.js
```diff
@@ -223,13 +223,17 @@ let showWireTemplate = {
 
 let toggleMenuTemplate = {
   i18n: 'menuShowHide',
+  type: 'checkbox',
+  checked: !init.restore('hideMenu', false),
```
lipis: let's rename the variable to `showMenu` isntead and have the `true` as default

ConorIA: By golly, you've turned this whole thing on its head! :wink:

electron/main.js
```diff
@@ -165,7 +165,7 @@ function showMainWindow() {
     'height': config.DEFAULT_HEIGHT_MAIN,
     'minWidth': config.MIN_WIDTH_MAIN,
     'minHeight': config.MIN_HEIGHT_MAIN,
-    'autoHideMenuBar': false,
+    'autoHideMenuBar': init.restore('hideMenu', false),
```
lipis: `showMenu`

ConorIA: Can it be `init.restore(!'showMenu', false)` or does it have to be `!init.restore('showMenu', true)`?

lipis:
Will merge after tests and update the translations

lipis:
Thank you @ConorIA 🍻

ConorIA:
Happy to do it! :smile:

PR #525, opened 2017-03-21
Update windows deploy scripts

mythsunwind:
* Both production and internal script now work the same. The s3 internal script does not upload RELEASES file directly. Instead this is now done by the deploy script (like on production).
* Fixed issue with environment variables when using client instead of resource client object on boto3 (tested with internal already)

### DIFF
--- a/bin/win-prod-s3-deploy.py
+++ b/bin/win-prod-s3-deploy.py
@@ -40,7 +40,7 @@ NEW_EXE_KEY = S3_PATH + NEW_EXE
 OLD_RELEASE_KEY = S3_PATH + 'RELEASES'
 OLD_EXE_KEY = S3_PATH + 'WireSetup.exe'
 
-s3 = boto3.resource('s3');
+s3 = boto3.resource('s3')
 
 s3.Object(BUCKET, OLD_RELEASE_KEY).delete()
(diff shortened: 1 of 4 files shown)


bin/win-prod-s3-deploy.py
```diff
@@ -40,20 +40,16 @@
 OLD_RELEASE_KEY = S3_PATH + 'RELEASES'
 OLD_EXE_KEY = S3_PATH + 'WireSetup.exe'
 
-client = boto3.client(
-  's3',
-  aws_access_key_id=os.environ.get('AWS_ACCESS_KEY'),
-  aws_secret_access_key=os.environ.get('AWS_SECRET_ACCESS_KEY')
-)
+s3 = boto3.resource('s3');
```
lipis: `;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;` :D

bin/win-prod-s3-deploy.py
```diff
@@ -40,20 +40,16 @@
...
 
-client.delete_object(Bucket=BUCKET, Key=OLD_RELEASE_KEY)
+s3.Object(BUCKET,OLD_RELEASE_KEY).delete()
```
lipis: space after `,`

PR #540, opened 2017-03-29
Open support in a new Window

lipis:


gregor:
Please also update the URL to the support website to not include locale and stuff after the page number

PR #541, opened 2017-03-29
Add certificate pinning with jsrsasign

ffflorian:


electron/main.js
```diff
@@ -192,15 +192,23 @@ function showMainWindow() {
...
-      cb(-2);
-    } else {
-      cb(-3);
+    if (certutils.hostnameShouldBePinned(hostname)) {
+      const pinningResult = certutils.verifyPinning(hostname, certificate);
+      if (pinningResult.verifiedFingerprints === false) {
+        console.error(`Certutils could not verify the certificate fingerprint(s) for ${hostname}.`);
+        return cb(-2);
+      }
+      if (pinningResult.verifiedIssuerRootPubkeys === false) {
```
lipis:
```js
if (!pinningResult.verifiedIssuerRootPubkeys) {
```

electron/main.js
```diff
@@ -192,15 +192,23 @@ function showMainWindow() {
...
     }
 
-    if (certutils.hostnameShouldBePinned(hostname) && !(certutils.verifyPinning(hostname, certificate))) {
-      console.error(`Certutils could not verify certificate pinning for ${hostname}.`);
-      cb(-2);
-    } else {
-      cb(-3);
+    if (certutils.hostnameShouldBePinned(hostname)) {
+      const pinningResult = certutils.verifyPinning(hostname, certificate);
+      if (pinningResult.verifiedFingerprints === false) {
```
lipis:
```js
if (!pinningResult. verifiedFingerprints) {
```

electron/main.js
```diff
@@ -192,15 +192,23 @@ function showMainWindow() {
     const {hostname = '', certificate = {}, error} = request;
 
     if (typeof error !== 'undefined') {
+      console.error(error);
```
lipis:
```js
console.error('setCertificateVerifyProc', error);
```

electron/main.js
ffflorian: Right now I'm setting `verifiedFingerprints` to `undefined` if no checkable fingerprints are found. Your requested change would mean that this line always fails, if no checkable fingerprints are found - this is not our desired result.

I specifically chose `undefined` because it defines an unknown state - because when no fingerprints where found for checking, `verifyPinning()` cannot tell if it can verify or not.

electron/html/certificate-error.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<!-- Wire, Copyright (C) 2016 Wire Swiss GmbH -->
```
gregor: It's 2017

electron/html/certificate-error.html
```diff
@@ -0,0 +1,35 @@
...
+        color: #fff;
+      }
+      a, a:hover, a:active, a:visited {
+        color: #fff;
+      }
+    </style>
+  </head>
+  <body ondragstart="return false;" ondrop="return false;">
+    <div id="errormessage">
+      Your connection is not private. <a href="https://wearezeta.zendesk.com/knowledge/articles/115002284109/en-us?brand_id=162184" target="_blank">Learn more</a>
```
gregor: I guess we should localize this. Not sure if we are set up to do on this page though. The link you added also is not the public one. Should be: https://support.wire.com/hc/articles/115002284109

electron/js/certutils.js
```diff
@@ -20,58 +20,89 @@
...
+  {
+    url: 'www.wire.com',
+    publicKeyInfo: {
+      algorithmID: ALGORITHM_RSA,
+      algorithmParam: null,
+      fingerprints: [MAIN_FP],
+    },
+  },
+  {
+    url: 'prod-nginz-https.wire.com',
```
gregor: What happens on our internal environments? And what about app.wire.com?

ffflorian: The internal environments are not pinned. app.wire.com is defined in `config.PROD_URL`, but I will change this, too.

electron/html/certificate-error.html
ffflorian: > I guess we should localize this

Then we should also localize [splash.html](https://github.com/wireapp/wire-desktop/blob/master/electron/html/splash.html#L43).

lipis: We can do that later as we didn't have anything before the [about changes](https://github.com/wireapp/wire-desktop/blob/master/electron/js/about.js#L25-L31) last week..

electron/main.js
```diff
@@ -189,17 +190,26 @@ function showMainWindow() {
   }
 
   main.webContents.session.setCertificateVerifyProc((request, cb) => {
-    const {hostname = '', certificate: {data: cert = ''} = {}, error} = request;
+    const {hostname = '', certificate = {}, error} = request;
 
     if (typeof error !== 'undefined') {
+      console.error(setCertificateVerifyProc, error);
```
lipis: `'setCertificateVerifyProc'`

lipis: just to find easier where this error is coming from

ffflorian:
So we will not translate the page for now, because `nodeIntegration: false` is set in the main browser window and for now there's no usable way to enable it at runtime. See https://github.com/electron/electron/issues/5113 and https://github.com/wireapp/wire-desktop/pull/541#discussion_r109173141.

Issue #542, opened 2017-03-29
How to verify integrity and authenticity of downloaded application?


dosvarog:
Hi, I am looking all over the web page, but I cannot find SHA256 (or similar) sums for downloaded applications so that I would be able to verify their integrity and authenticity (on Linux, which I use, that would be with tools like `sha256sum` and `gpg` or similar). Something like Linux Mint has on their web page (and let us not forget that hackers hacked their servers and planted their version of ISO file) or Ubuntu Linux on theirs (https://linuxmint.com/verify.php and https://www.ubuntu.com/download/how-to-verify).

So, how can I verify that I have downloaded correct version of application, i.e. that I have not downloaded version that was in any way manipulated by third side (and planted on a web page in a download section)?

raphaelrobert:
Go to wire.com/download and hover on 'Details'.

maximbaz:
I wonder if it's possible to add a PGP-signed file containing checksums directly to the github releases page. Have a look [here](https://github.com/syncthing/syncthing/releases/) for an example, note the file `sha256sum.txt.asc`. 

By the way, there is no `Details` to hover for source code or Linux releases on the https://wire.com/download page.

dosvarog:
@maximbaz Nice idea!
@raphaelrobert Yes, thank you, I see it (for Android and Windows), but as @maximbaz said, there is no checksum for either .deb or AppImage (I know it says on the page that Linux build is experimental, but still, checksums would be very nice). I just saw someone [added a commit](https://github.com/wireapp/wire-desktop/pull/543/commits) for that. I guess, now it only needs to be put on a web site.

raphaelrobert:
Thanks, we are looking to improve things for linux.

3n-mb:
Actually, there are two questions:
 - How to verify this electron app?
 - How to verify crypto and UI code that is downloaded by electron every time wire app is started?

In main.js, starting with [line 122](https://github.com/wireapp/wire-desktop/blob/master/electron/main.js#L122) we see that code that handles all of your security is downloaded every time wire app is run.

This code can be new every time. There is no point in checking electron's app integrity, when it dutifully executes *anything* that comes from wire's server, with `nodeIntegration = true` !

[Issue 17](https://github.com/wireapp/wire-desktop/issues/17) still stands.

If someone hacks wire's server, they can serve code, which in electron, with `nodeIntegration = true`, owns you. Running the same malicious code in browser, with full browser sandbox is safer.

[...]

PR #543, opened 2017-03-30
Add SHA256SUMS to deploy work flow (Addresses #542)

ConorIA:
I saw #542 and thought, "Hey, why not?".
If you folks like this small addition, please make sure to check the s3 deploy code. I am one of those "R, never Python" guys (well, not _never_.).


ConorIA:
This seems to completely conflict with #550, so I'll close this.

lipis:
we will improve the signing and/or the hashes.. so stay tuned.

PR #549, opened 2017-04-03
Use libsodium-neon

ffflorian:


electron/js/preload.js
```diff
@@ -137,6 +137,11 @@ ipcRenderer.once('webapp-loaded', function(sender, config) {
     });
   }
   // else we are on /auth
+  try {
+    Object.assign(window.sodium, require('libsodium-neon'));
+  } catch (err) {
+    // fallback to libsodium.js
```
bennyn: Maybe we should log something here (if possible) in our `console.log` file!?

ffflorian: We can, but I thought the idea was to silently fail. :)

PR #550, opened 2017-04-03
Rename package to wire-desktop

lipis:


package.json
```diff
@@ -24,14 +24,6 @@
...
       "output": "wrap/dist"
-    },
-    "linux" : {
-      "afterInstall" : "bin/deb/after-install.tpl",
-      "afterRemove" : "bin/deb/after-remove.tpl",
-      "asar": false,
-      "category": "Network",
-      "depends": ["libappindicator1", "libasound2", "libgconf-2-4", "libnotify-bin", "libnss3", "libxss1"],
-      "target": ["AppImage", "deb"]
     }
```
herrmannplatz: we can remove "grunt-shell": "2.1.0",

ConorIA:
I expect that this will break the packaging strategy I use on the AUR. Is there any chance you could add a "dir" build option? ie without making the deb or AppImage

mythsunwind:
Hi ConorIA, I checked your `PKGBUILD` file. As far as I understand you use the `build --linux --{arch} --dir` to get both `/wrap/dist/linux-unpacked/` and `/wrap/dist/linux-ia32-unpacked/`. These directories will be generated still when running grunt `linux-prod` on this branch, it just will take longer and builds both architectures.

But maybe I can add a special `dir` build option for you...

ConorIA:
Thank you @mythsunwind. I'll experiment a little tonight. I might just have to run some sed commands before starting the build.

mythsunwind:
@ConorIA I added another grunt task. You can now run `grunt clean:linux update-keys release-prod electronbuilder:linux_dir` to only create the unpacked directories. Please let me know if you need additional changes.

ConorIA:
@mythsunwind, thank you! I will play with the PKGBUILD this afternoon. Will this spit out dies for both ia32 and x64, or does builder detect the arch automatically?

mythsunwind:
@ConorIA It will spit out both and currently does not detect the arch. I assume it would be better to only build the current architecture when using `linux_dir`. I could add a "if" for this.

ConorIA:
That _would_ be useful.

This is how something like that might look on Arch-based distros.

```
if [ $CARCH == 'x86_64' ]; then
  ... x64
elif [ $CARCH == 'i686' ]; then
  ... ia32
else
  echo "Unsupported architecture"; exit 1;
fi
```

I am not sure if all Linux distros use the same two architecture codes.

mythsunwind:
You now have two commands:
`grunt clean:linux update-keys release-prod electronbuilder:linux_ia32`
and 
`grunt clean:linux update-keys release-prod electronbuilder:linux_x64`

Both only create the unpacked directories (no AppImage or deb) for you to use in `PKGBUILD`.

package.json
mythsunwind: done

ConorIA:
@mythsunwind, thank you for your changes to accommodate the AUR package. It seems that the list of different Linux build options is getting a little long, so I have come up with something that you might consider using: 

```
      linux_other:
        options:
          arch: "#{process.env.WIRE_ARCH}"
          productName: 'wire-desktop'
          targets: ["#{process.env.WIRE_TARGET}"]
          linux:
            fpm: ['--name', 'wire-desktop']
            executableName: 'wire-desktop'
```

This could replace the two other options, and would be called as such (tested working):
```
WIRE_ARCH="x64" WIRE_TARGET="dir" grunt electronbuilder:linux_other
WIRE_ARCH="ia32" WIRE_TARGET="dir" grunt electronbuilder:linux_other
```

As an added bonus, this keeps things modular so that if a builder wants to build to another target, they can. e.g. 
```
WIRE_ARCH="x64" WIRE_TARGET="rpm" grunt electronbuilder:linux_other
```

ffflorian:
I just tested this and it works perfect! All the data is still there. :+1:

mythsunwind:
@ConorIA nice idea, but how about using grunt option flags instead and avoid using environment variables? This would make something like `grunt --target=rpm --arch=x64 electronbuilder:linux_other` possible.

I would say, let's go with the current way and create a new ticket for putting grunt option flags to the roadmap. The main purpose of this PR is to get compliant with official Debian/Ubuntu repositories while not breaking the AUR packaging.

Let me know if you need more help for updating the PKGBUILD. We don't want to merge this and break everything on arch linux. The binary is now located at `/opt/wire-desktop/wire-desktop` instead of `/opt/Wire/Wire`. I hope this works for you as well!?

ConorIA:
@mythsunwind, I updated the `wire-desktop-git` PKGBUILD locally yesterday. Everything is working. I'll update it on the AUR once this is merged. The main PKGBUILD will be updated when you folks make a new release. If you are ready to go, merge away!

Issue #575, opened 2017-04-14
Ubuntu 17.04: tray icon disappears


sebastianbauer:
After updating Ubuntu to 17.04 (on 2 PCs) there is no longer the wire tray icon available.

This was a problem earlier. See #22.

`libappindicator1` is installed.

ConorIA:
Dupe of #574.

Issue #574, opened 2017-04-13
No system tray icon with Ubuntu 17.04


smares:
After upgrading my Ubuntu 16.10 installation to 17.04, the Wire icon disappeared from the system tray. Using Wire 2.12.2729.

ConorIA:
Unity is finicky. Please see if this works for you: http://www.webupd8.org/2017/04/fix-appindicator-not-working-for.html

Note that this is nor a Wire-specific bug, as it seems to affect all Electron-based apps.

PR #576, opened 2017-04-15
Workaround Ubuntu 17.04 tray bug

ConorIA:
This is an untested workaround for #574 and #575. I have used the console.log, and the property does seem to be set. (though in my case I was changing `KDE` to `Unity`).

If one of you has a working Zesty setup, please test. I've build a [`.deb` (x64)](https://cloud.conr.ca/s/RXfiBV3NX8HZSDf/download) to save you from building. If no one has a 17.04 install, I can ask the reporter take it for a spin.

If this does work, we should let the folks over at https://github.com/electron/electron/issues/9046 know.


electron/main.js
```diff
@@ -89,6 +89,15 @@ if (process.platform !== 'darwin') {
 }
 
 ///////////////////////////////////////////////////////////////////////////////
+// Fix indicator icon on Unity
+// Source: https://bugs.launchpad.net/ubuntu/+bug/1559249
+///////////////////////////////////////////////////////////////////////////////
+const de = process.env.XDG_CURRENT_DESKTOP
+if (de.includes("Unity")) {
```
ConorIA: There might be a better way to do this test. We need to change from `Unity:Unity7` to `Unity`.

electron/main.js
```diff
@@ -89,6 +89,15 @@ if (process.platform !== 'darwin') {
 }
 
 ///////////////////////////////////////////////////////////////////////////////
+// Fix indicator icon on Unity
+// Source: https://bugs.launchpad.net/ubuntu/+bug/1559249
+///////////////////////////////////////////////////////////////////////////////
+const de = process.env.XDG_CURRENT_DESKTOP;
+if (de.includes('Unity')) {
```
ConorIA: There might be a better way to do this test. We need to change from Unity:Unity7 to Unity.

ConorIA:
Went in with Virtualbox.
Success!

screenshot_20170415_120820

electron/main.js
lipis: Just skip using the variable `de` and instead `process.env.XDG_CURRENT_DESKTOP.includes..`

lipis:
Looks good to me.. @mythsunwind wdyt?

lipis:
and @ffflorian :)

mythsunwind:
To me too. Let's merge it before the new release.

lipis:
🍻 🇪🇺 🍺 🇨🇦

PR #599, opened 2017-04-20
Prevent duplicate tray icon on update (WEBAPP-3854)

gregor:


electron/main.js
```diff
@@ -108,6 +100,21 @@ if (process.platform === 'win32') {
...
+    // Stop further execution on update to prevent second tray icon
+    return;
+  }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// Fix indicator icon on Unity
+// Source: https://bugs.launchpad.net/ubuntu/+bug/1559249
+///////////////////////////////////////////////////////////////////////////////
+if (process.platform === 'linux') {
```
bennyn: Instead of having all these `if`'s here and having a comment with `Fix indicator icon on Unity` we should consider having a function called `fixIndicatorIconOnUnity` or something like that. :)

bennyn: What is also a nice coding pattern is to use functions over conditions. 

**Example:**

```javascript
function isUnity() {
  if (process.env.XDG_CURRENT_DESKTOP && process.env.XDG_CURRENT_DESKTOP.includes('Unity')) {
    return true;
  }
  return false;
}

if (isUnity()) {
  process.env.XDG_CURRENT_DESKTOP = 'Unity';
}
```

PR #603, opened 2017-04-22
Fix debian pre/post-install scripts

### LINKED ISSUE #602: [Linux] bad link in /usr/local/bin
When instaling wire-desktop on Ubuntu 16.04 from the PPA, the setup creates the following link in `/usr/local/bin`: `<%= executable %> -> /opt/<%= productFilename %>/<%= executable %>`.

It looks like the placeholders are not replaced with the actual values.

ConorIA:
A [change on the `electron-builder` side](https://github.com/electron-userland/electron-builder/commit/fe137fc221f46683433cc7dc86984caeab281d84) broke the `after-install.tpl` and `after-remove.tpl` scripts. This resolves #602, but the bad links might still exist for some users.

### DIFF
--- a/bin/deb/after-install.tpl
+++ b/bin/deb/after-install.tpl
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 # Link to the binary
-ln -sf '/opt/<%= productFilename %>/<%= executable %>' '/usr/local/bin/<%= executable %>'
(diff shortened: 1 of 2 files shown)


ConorIA:
I should have said: I noticed that the old `after-remove. tpl` was trying to remove from `/usr/local/bin`, but the links are actually in `/usr/bin`. Perhaps we should make `after-install.tpl` try to remove old invalid links too?

PR #610, opened 2017-04-25
Use libsodium-neon

ffflorian:


electron/js/preload.js
```diff
@@ -137,6 +137,13 @@ ipcRenderer.once('webapp-loaded', function(sender, config) {
     });
   }
   // else we are on /auth
+  try {
+    Object.assign(window.sodium, require('libsodium-neon'));
+    console.log('Using libsodium-neon.');
+  } catch (err) {
```
bennyn: `err` -> `error`

electron/js/preload.js
```diff
@@ -137,6 +137,13 @@ ipcRenderer.once('webapp-loaded', function(sender, config) {
...
+    // fallback to libsodium.js
```
bennyn: Comment is not necessary. If you want to keep this information, then just put it into the log statement. 😃 

```javascript
 console.log('Failed loading "libsodium-neon", falling back to "libsodium.js".', error);
```

PR #678, opened 2017-06-08
Add Prettier

lipis:
[styled with prettier](https://github.com/prettier/prettier)

Related: https://github.com/wireapp/wire-webapp/issues/1347


.eslintrc.yml
```diff
@@ -39,5 +39,4 @@ rules:
   ]
   sort-keys: [2, asc, {caseSensitive: true, natural: true}]
   sort-vars: 2
-  spaced-comment: [2, always, {exceptions: ['#']}]
```
gregor: Shouldn't we rather add the `/` to solve the issue and still have a rule for this? -> ``spaced-comment: [2, always, {exceptions: ['#', '/']}]`

PR #718, opened 2017-06-25
Removed init lib in favor of settings lib

pwnsdx:
- Less usage of disk
- Only one read of the config file is required
- Only one async write required
- Cleaner class


electron/js/lib/ConfigurationPersistence.js
```diff
@@ -0,0 +1,83 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const app = require('electron').app || require('electron').remote.app;
```
bennyn: Do we need the `||`?

electron/js/lib/ConfigurationPersistence.js
```diff
@@ -0,0 +1,83 @@
...
+    const value = global._ConfigurationPersistence[name];
+    return (typeof value !== 'undefined' ? value : defaultValue);
+  }
+
+  _saveToFile() {
+    return new Promise((resolve, reject) => {
+      const datasInJSON = JSON.stringify(global._ConfigurationPersistence);
+      this.debug('Saving datas to persistent storage: %o', datasInJSON);
+      fs.writeFile(INIT_JSON, datasInJSON, 'utf8', (err, data) => {
+        if (err) {
```
bennyn: Sometimes it's called `err`, sometimes it's called `e` and sometimes `error`. Let's stick to the most explicit one which is `error`.

electron/js/lib/ConfigurationPersistence.js
pwnsdx: No idea but it was here originally so

electron/main.js
```diff
@@ -323,15 +323,23 @@ function showMainWindow() {
...
     if (!main.isFullScreen()) {
       init.save('bounds', main.getBounds());
     }
 
     if (!quitting) {
       event.preventDefault();
       main.hide();
+
+    } else {
+
```
bennyn: Why there is an empty line?

electron/js/lib/ConfigurationPersistence.js
```diff
@@ -0,0 +1,83 @@
...
+const app = require('electron').app || require('electron').remote.app;
+const INIT_JSON = path.join(app.getPath('userData'), 'init.json');
+const debug = require('debug');
+
+class ConfigurationPersistence {
+
+  constructor() {
+    this.debug = debug('ConfigurationPersistence');
+
+    // Get content of init.json
```
bennyn: No need for having this comment. It's clear by the code.

electron/main.js
```diff
@@ -323,15 +323,23 @@ function showMainWindow() {
...
+      // Save configuration to disk
```
lipis: no need for this comment.. pretty clear

electron/js/menu/system.js
```diff
@@ -24,9 +24,9 @@ const autoLaunch = require('auto-launch');
 const launchCmd = (process.env.APPIMAGE != null) ? process.env.APPIMAGE : process.execPath;
 
 const config = require('./../config');
-const init = require('./../lib/init');
 const locale = require('./../../locale/locale');
 const windowManager = require('./../window-manager');
+const init = require('./../lib/ConfigurationPersistence');
```
lipis: they were sorted alphabetically.. so let's keep it that way

pwnsdx: Already fixed

lipis:
init was a bit general.. maybe we could name it `settings` instead of ConfigurationPersistence?

electron/js/lib/ConfigurationPersistence.js
pwnsdx: Apparently used so it can be compatible with both main process and renderer

lipis: yes we need to keep it

PR #717, opened 2017-06-25
Cleanup

pwnsdx:
- Using const instead of let for "argv"
- Separated defines
- Use ` ${}` instead of ' '+
- Added comments

PR #726, opened 2017-06-26
feat: redux store

herrmannplatz:


.eslintrc.yml
```diff
@@ -1,8 +1,11 @@
 ---
+  parserOptions:
+    sourceType: module
+    ecmaVersion: 8
```
pwnsdx: https://github.com/wireapp/wire-desktop/pull/717/commits/d69520e3fb94720a6cb1ba7feedb920726dd3886 :)...

herrmannplatz: 👍

electron/renderer/src/index.js
```diff
@@ -1,7 +1,26 @@
...
+const persistedState = loadState()
+
+const store = createStore(
+  appStore,
+  persistedState
+)
+
+store.subscribe(() => {
+  saveState({
+    accounts: store.getState().accounts
```
Yserz: Annotation: Here we could use reselect or lodashs memoize to reduce the writes to the local storage by only calling the saveState when the state for the account actually changed.

PR #730, opened 2017-06-26
feat: basic react ui

herrmannplatz:


electron/renderer/src/components/Webview.js
```diff
@@ -0,0 +1,35 @@
...
+    constructor(props) {
+        super(props)
+
+        this._onPageTitleUpdated = this._onPageTitleUpdated.bind(this)
+    }
+
+    componentDidMount() {
+        const {src, partition} = this.props
+
+        // set unknown props
```
Yserz: I think that comment is outdated now :)

electron/renderer/src/containers/TeamsSidebar.js
```diff
@@ -0,0 +1,28 @@
...
+    onAddAccountClick: () => {
+      dispatch(addAccountWithSession())
+    },
+    onSwitchAccountClick: (id) => {
+      dispatch(switchAccount(id))
+    }
+  }
+}
+
+const TeamsSidebar = connect(
```
Yserz: WDYT about moving the connect to the Sidebar component since it has already team specific elements like team icon and variable names like `account`? Do you prefer to have it separated to have different types of sidebars?

electron/renderer/src/static/webview-preload.js
```diff
@@ -0,0 +1,118 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+'use strict';
+
+alert('foo')
```
Yserz: chore

PR #751, opened 2017-07-05
feat: context menu

herrmannplatz:


electron/renderer/src/components/Sidebar.js
```diff
@@ -53,14 +54,21 @@ const Sidebar = ({
...
+              <path d="M0 6.125v1.75h6.125V14h1.75V7.875H14v-1.75H7.875V0h-1.75v6.125" fillRule="evenodd"/>
+            </svg>
+          </div>
         </div>
-      </div>
+      </ContextMenuTrigger>
     }
+
+    <ContextMenu id="account">
+      <ContextMenuItem onClick={() => window.open('https://teams.wire.com')}>Create Team</ContextMenuItem>
```
gregor: Guess we should defined this in a variable long term - somewhat environment dependent.

herrmannplatz: yes

PR #757, opened 2017-07-05
Use a switch instead of ifs for web-contents-created event

pwnsdx:


electron/main.js
```diff
@@ -406,7 +406,8 @@ class ElectronWrapperInit {
 
     app.on('web-contents-created', (event, contents) => {
 
-      if (contents.getType() === 'window') {
+      switch(contents.getType()) {
+        case 'window':
         contents.on('will-attach-webview', (e, webPreferences, params) => {
```
gregor: Missing indentation for case content

PR #769, opened 2017-07-08
Moved getBaseUrl function in a const + Added BrowserWindowInit

pwnsdx:


electron/main.js
```diff
@@ -475,6 +475,53 @@ class ElectronWrapperInit {
       }
     });
   }
-}
+};
+
+class BrowserWindowInit {
```
pwnsdx: No, in a later PR.

PR #798, opened 2017-07-12
fix: handle click on context menu + check for maximum accounts limit

herrmannplatz:


gregor:
Doesn't this break the changes we made yesterday when we introduced  `addAccount(true)`?

herrmannplatz:
you mean the delete account issue?

gregor:
It doesn't. Thanks for clarifying.

PR #819, opened 2017-07-21
[WIP] Add .spec file (adds RPMs to release offering)

ConorIA:
I spent a lot of hours (more than I should have) this week learning to set up a build machine and repo on my [lab server](https://gitlab.com/ConorIA/wire-builds). Then @1Claude comes along and links me to https://copr.fedorainfracloud.org, and makes all that hard work go to waste! 

Here we have the tools to create real, working RPM builds, not the nasty things that `electron-builder` spits out. The platform also makes yum repos automatically, I am happy to monitor the builds, as long as users like @1Claude and others are happy to test.

## What else needs to be done:
- [x] Make your release script bumps the version number in the `.spec` when there is a new release (I can have a look, but I'm sure you folks will be faster)*
- [ ] Change the current settings from my fork to the official repo and master branch (I can do this once it is merged)
- [ ] Add a GitHub webhook to start a rebuild when a new version is released (you folks need to take care of this one. I can send the webhook URL over Wire)

## Advantages of copr:
- Rebuild for all versions of Fedora, CentOS, etc.
- Offloads all the build work to the Fedora infrastructure. 

## Advantages of my builds on my server:
- GPG signed (by me, of course)
- Google API keys are set, so contacts import should work. (can't seem to set these on copr)

\* right now the implementation I've got here will bump the spec for every tag/release. How could we make it stick to "official" releases only?


ConorIA:
Come to think of it, the version bump probably shouldn't happen in the `Gruntfile`, but should instead become part of the release process in the `bin/` folder.

mythsunwind:
If you check wireapp/wire-desktop/pull/817 you can see that we currently work on streamlining the build process and generation the build numbers differently. This might break your PR.

ConorIA:
Hmm. When you merge #817, I'll double check and make the necessary adjustments. 👍

mythsunwind:
Maybe you can move the .desktop and .spec file in a directory called `bin/rpm`, then it is following the same structure as `bin/deb`.

Instead of the Gruntfile changes, I could also think about a bash script in `bin/rpm` that will update the version number in the .spec file after build.

ConorIA:
I tried building with the spec in a subdirectory, but the builder couldn't find the files because it looks for them from within the subdirectory. There is a way around this in the `%prep` macro, but I haven't been able to figure it out. I can experiment more tonight. If any user who is in the know stumbles upon this... Help!

The `.desktop` file can be moved, but I really do think it is  a resource. I'll switch to using it in my Arch builds instead of gendesk. If I could figure out how to make `electron-builder` use it, I'd submit another PR! This file is to spec and thorough. I'd like to see it localized too to support the languages Wire supports.

I'll take a look again tonight.

ConorIA:
I tried a bunch of little tweaks today (as one can see from the commit log). 
Copr is super frustrating because there is so little documentation or anything to deal with errors. I'm basically back to where we were back in July. I cannot find a way for Copr to build from a subdirectory, because it can't find the correct build folder when it downloads from a git tarball. I can't redefine the version in the specfile, because Copr can't import it. 

Any recommendations around where to bump the version number? Ideally it would be bumped automatically along with `electron/package.json` and `info.json`, like in https://github.com/wireapp/wire-desktop/commit/d262e31f48c63e5dae9fdd5f6a603c2ade8a59be.

ConorIA:
This is too ugly. I think I'll just maintain my https://gitlab.com/ConorIA/wire-builds/ repo and have the pipelines send a source RPM to copr.

mythsunwind:
Oh, no. Sad to hear. We changed our build method some weeks ago. The major.minor version is now only changed in info.json. The build id (last number of the version string) is set by the environment variable BUILD_NUMBER.

PR #821, opened 2017-07-21
Create sha256sum.txt and sign it in build process

mythsunwind:


bin/linux-checksums.sh
```diff
@@ -0,0 +1,81 @@
...
+
+if [ "${SHRED_STATUS}" == "unavailable" ]; then
+  _log "Info: shred not found. Using insecure way of deleting."
+else
+  find "${GPG_TEMP_DIR}" -type f -exec shred {} \;
+fi
+
+rm -rf "${GPG_TEMP_DIR}"
+
+_log "Done"
```
gregor: Please add empty last line

PR #830, opened 2017-07-24
fix: Handle umlaut in og:image (WEBAPP-4156)

herrmannplatz:


lipis:
@herrmannplatz I think we should close this one.. right?!

gregor:
@lipis  This is supposed to fix a fairly annoying bug. I will take a look and try to get this into a mergable shape.

PR #835, opened 2017-07-26
Upload builds as drafts

mythsunwind:
tl;dr: enhances continuous deployment pipeline

The build job will create a release draft in GitHub and uploads certain build artifacts (deb package, AppImage, gpg signed checksums, Windows executable, macOS pkg) directly after the build. These release drafts can only be seen by users with write access on the repository. The draft can be used to do the final release or discarded when not needed.


bin/github_draft.py
```diff
@@ -0,0 +1,89 @@
...
+          if response.status_code in [200, 204]:
+            print 'Draft deleted!'
+          else:
+            print 'Delete failed with %s :(' % (response.status_code)
+            print response.text
+          exit(1)
+  else:
+    print 'Error %s :(' % (response.status_code)
+    print response.text
+    exit(1)
```
gregor: Please add empty line EOF

mythsunwind: Done

PR #934, opened 2017-09-13
refactor: Replaced global logger with logger instance

bennyn:


electron/main.js
```diff
@@ -391,25 +382,22 @@ app.on('ready', function() {
...
+  contents.map((file) => path.join(LOG_DIR, file, config.CONSOLE_LOG))
+    .filter((file) => {
+      try {
+        return fs.statSync(file).isFile();
+      } catch(error) {
+        return undefined;
       }
+    })
+    .forEach((file) => {
+      if (file.endsWith('.log')) fs.renameSync(file, file.replace('.log', '.old'));
```
gregor: Can we stay without the single line if statements?

Issue #957, opened 2017-09-19
Gnome Desktop Issues: Removal of legacy status icons panel, wire app pin to panel 


HeritageRD:
Wire Version: 2.15.2751
Wire for Web Version: 2017.08.31.1421
OS: Arch Linux 64-bit
Do you have an antivirus software installed: No
Use Case: Non-technical office environment

What steps will reproduce the problem?
1. As of Gnome Desktop version 3.26, the legacy system tray is removed preventing Gnome Wire users from interacting with the Wire app once the primary window is closed (despite the app continuing to run in the background). A 3rd-party extension (the semi-buggy top-icons addon) is available to restore this functionality, however in the next version of the GTK toolkit, the underlying notification system is to be removed rendering this functionality completely lost.   
2. In an attempt to avoid this issue, I tried to utilize the Wire Icon in Gnome's favourites panels. However, this icon cannot be pinned to the favourites panel or pinned utilizing the dash to dock extension or dash to panel extension.   

What is the expected result?
The Wire app should  remain visible in the Gnome favourites panel to access the app in absence of a system tray (or equivocal extension) regardless of the presence of active windows. This app should be pin-able to the panel.

Other thoughts:
To my understanding the issue is coming from an unfortunate intersection between the behaviour of electron apps and Gnome's "vision" for the future. Any thoughts on a possible course of action?  

Thanks for the development of this app!

HeritageRD:
ConorIA,

Thanks for looking into this. I ran the set command as you requested. Attached is the output. I did not see anything which directly stated the version of gnome desktop environment. If you need additional information, let me know. 

[set.txt](https://github.com/wireapp/wire-desktop/files/1329673/set.txt)

davidhaller:
Ubuntu 17.10 also uses GNOME 3.26, it comes with an AppIndicator extension that uses some different api then the legacy tray icons. Wire does not work with it, but other apps (e.g. Skype) do.

EmPeWe:
> In the meantime, I suggest using the window close button should quit Wire when running on gnome. This can be detected by reading the $DESKTOP_SESSION env variable.

The patch from ff6cc15a2734f65a031231ef2117e31354f466a2 is far better than just quitting wire. Maybe it's possible to make this a user's decision within the wire preferences?

PR #970, opened 2017-09-23
Do not hide window when closing on Gnome (#957)

ConorIA:
Ideally the function could check whether Gnome was >= 3.26, but I am not sure if there is an environment variable for that. Maybe @HeritageRD wouldn't mind checking whether there is anything apparent when they run `set` from a terminal.

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -149,7 +149,7 @@ if (process.platform === 'linux') {
 ///////////////////////////////////////////////////////////////////////////////
 // Make sure not to close the window to the tray on Gnome
 ///////////////////////////////////////////////////////////////////////////////
-function closeButton() {
+function onCloseMain() {
   if (process.platform === 'linux' && process.env.XDG_CURRENT_DESKTOP && process.env.XDG_CURRENT_DESKTOP.includes('GNOME')) {
     main.minimize();
   } else {
@@ -310,11 +310,11 @@ function showMainWindow() {
 
       if (isFullScreen) {
         main.once('leave-full-screen', () => {
-          closeButton();
+          onCloseMain();
         });
         main.setFullScreen(false);
       } else {
-        closeButton();
+        onCloseMain();
       }
     }
   });
--- a/electron/main.js
+++ b/electron/main.js
@@ -147,6 +147,17 @@ if (process.platform === 'linux') {
 }
 
 ///////////////////////////////////////////////////////////////////////////////
+// Make sure not to close the window to the tray on Gnome
+///////////////////////////////////////////////////////////////////////////////
+function closeButton() {
+  if (process.platform === 'linux' && process.env.XDG_CURRENT_DESKTOP && process.env.XDG_CURRENT_DESKTOP.includes('GNOME')) {
+    main.minimize();
+  } else {
+    main.hide();
+  }
+}
+
+///////////////////////////////////////////////////////////////////////////////
 // IPC events
 ///////////////////////////////////////////////////////////////////////////////
 ipcMain.once('webapp-version', (event, version) => {
@@ -299,11 +310,11 @@ function showMainWindow() {
 
       if (isFullScreen) {
         main.once('leave-full-screen', () => {
-          main.hide();
+          closeButton();
         });
         main.setFullScreen(false);
       } else {
-        main.hide();
+        closeButton();
       }
     }
   });


electron/main.js
```diff
@@ -299,11 +310,11 @@ function showMainWindow() {
 
       if (isFullScreen) {
         main.once('leave-full-screen', () => {
-          main.hide();
+          closeButton();
         });
         main.setFullScreen(false);
       } else {
-        main.hide();
+        closeButton();
```
gregor: @ConorIA Thanks for taking a look at this. We are not overly excited to add to the list of exceptions for specific environments but I guess there is currently no way around in order to address the issue. Can I ask you to name the function a little more verbose those? Possibly `onCloseMain` as the function is somewhat unrelated to its current name.

ffflorian:
Hi @ConorIA,
thank you for the PR. Could you please explain your reasons to create this behaviour? I don't see a good reason to break the default behaviour to hide a window when you close it. This behaviour also exists in other apps with status icons like Telegram, Riot, Nextcloud, etc.

Furthermore, with a right-click on the window an click on *Minimize* or simply by pressing <kbd>Super</kbd>+<kbd>H</kbd> you can hide any window.

ConorIA:
@ffflorian, in #957 @heritagemd explained that in the latest Gnome version, there is no system tray. That means that a closed window disappears with no tray icon to restore it. The reasoning behind this PR is to have the window minimized, since that is the only way (to my limited knowledge) to keep the window accessible without re-launching.

gregor:
Fixes #957

mythsunwind:
Here are the official guidelines for the migration of status icons from Gnome:
https://wiki.gnome.org/Initiatives/StatusIconMigration

A long article explains the reasoning behind it:
https://blogs.gnome.org/aday/2017/08/31/status-icons-and-gnome/

The people from telegram desktop discussing it too: 
https://github.com/telegramdesktop/tdesktop/issues/3830

We already have native notifications, but lack a solution for the quit/close issue. Minimizing instead of closing is just a workaround and might not fix the issue described for Wire on their issue tracking list for some users. The issue is not that you cannot reopen it while running, but that you cannot re-access the app at all, after you once closed it when it was hidden. The steps are:

1. Open Wire on Gnome
2. Set "Launch Wire on Startup"
3. Hide window by clicking X 
4. Logout & Login
**Actual:** Wire is only in status icon area. If this area is removed in Gnome 3.26, the user is unable to bring wire to the foreground again.

The Gnome guidelines suggest to add a popup that ask the user to make the app quit or run in background. @ConorIA what do you think about this?

ConorIA:
@mythsunwind, I haven't used Gnome in some time now, but I think that a user could get back into Wire by attempting to launch a new instance. IIRC, there is a `main.show()` in the single instance stuff. So I guess, theoretically Wire _could_ run in the background. The popup sounds super clunky, but I guess it is one to run by your design team. 

Perhaps another option: instead of detecting Gnome, you could just add a "close to the status bar" option that is enabled by default. The option could thrn be disabled and greyed-out on Gnome.

jhasse:
It's possible to use an extension for tray icons or AppIndicators, therefore I don't think that this functionality should be disabled based on the `XDG_CURRENT_DESKTOP` environment variable alone.

ConorIA:
@jhasse the tray icon will still be there, the app just won't close to the tray. I guess then a "close to tray" option would be preferable for you?

jhasse:
I think Wire should behave in the same way on all DEs (preferebly even on all operation systems). So as it currently closes to the tray icon (not with an option), I prefer that.

chrisHRD:
@jhasse 
I have tried to do some research into the topic of tray icons. Here's what I have found. The issue is a bit systemic thanks to the Gnome developers' vision.
- Top icons extension is pausing development due to and API issue [OMGUbuntu Article](http://www.omgubuntu.co.uk/2017/09/top-icons-plus-development-paused) 
- Canonical, is moving towards the extension called KStatusNotifierItem/AppIndicatorSupport for their next version of Ubuntu based on Gnome. [OMGUbuntu Article](http://www.omgubuntu.co.uk/2017/08/ubuntu-sees-sense-will-support-indicator-applets-ubuntu-17-10)
- @mythsunwind's article written by the Gnome dev hints at a possible future with more reductions in the code supporting these kind of notifications.

If Gnome was not the default for so many major distributions, this would be less of an issue. But as it stands, it holds a strong market share in the Linux community and a future left up to third-party extensions is uncertain at best. I agree with @jhasse that uniformity is foundational to a cross-platform app's usability; however, if critical control methods are no longer available "natively" (at least on the Linux side), I wonder about stability in the long-run. From what I have seen KDE Plasma (2nd largest desktop environment) is not going to follow this trend, so perhaps making sure Wire works with KStatusNotifierItem/AppIndicator backend is a safe solution for now. Thoughts?

jhasse:
> so perhaps making sure Wire works with KStatusNotifierItem/AppIndicator backend is a safe solution for now. Thoughts?

Agreed. With Ubuntu's GNOME shipping the extension by default and all other DEs in mind, I think that no-tray-at-all is in the minority here (only vanilla GNOME).

Electron already added support for AppIndicators in GNOME, unfortunately under Ubuntu only: https://github.com/electron/electron/commit/dc43dc2a1316d5f96e1c3a9cf6079c4150214f66
There's an open issue about a proper implementation though: https://github.com/electron/electron/issues/10427

Issue #649, opened 2017-05-21
[Feature request] Windows portability


vatterspun:
Obviously Wire is already platform-portable, meaning it works on multiple operating systems.  The goal here is to be settings-portable, meaning it would work across Windows machines.

## Current behavior

Writes the program to appdata and writes settings to appdata\local and appdata\roaming.

## Requested behavior

1. Write program and all settings to the same folder (or at least the same root folder) e.g. c:\wire\ and c:\wire\settings
2. When the folder is moved to another folder (like x:\wire or c:\Users\Admin\Desktop) the program should launch with the same settings as before

## Why is this useful?

* Ease of encryption e.g. store inside a TrueCrypt / VeraCrypt volume

* USB drive - Someone who uses a public terminal at an internet cafe, computer lab, etc. 

* Cloud capability - An individual who has this stored on their cloud drive (Dropbox, Google Drive, etc.) will have the same experience/settings from one machine to another without a reinstall 

* Ease of backup - just copy the folder to a backup drive -- no hunting for the proper files inside appdata

## Who cares?

Several messaging tools offer a portable version for greater anonymity (it's easy to dump a flash drive).  Pidgin, two Tox clients, Telegram and Microsoft's Skype all have this capability.

## Possible issues
	
Going this route can make updating the client more complicated. When a new version comes out, users will have to avoid overwriting the old config file or lose their settings. As such, I recommend this as an optional setting (a command argument like "--portable") or separate release version.

smaragdus:
**1+** for portable version.
For example [Telegram](https://desktop.telegram.org/), [Isotoxin](http://www.isotoxin.im/) and all other [Tox clients](https://tox.chat/) offer portable versions or portable mode on Windows which facilitates usage and backup.

ConorIA:
For reference #397 #427

vatterspun:
> For reference #397 #427

Sorry @ConorIA I thought I ran a search before I posted.  My fault.

lipis:
We have to do something about #427 :)

janost:
+1, having a portable option would be nice.

ConorIA:
#427 has been merged, which means that one could install to a USB stick and run with the --portable flag. The only thing I am not sure about is whether the install location is customizable.

vatterspun:
@ConorIA Great news, thanks.

sergeevabc:
[...]

PR #982, opened 2017-09-27
Use electron-builder's portable env.

ConorIA:
Electron builder adds a `PORTABLE_EXECUTABLE_DIR` environment var when building a portable target. I did not have any luck when I tried to add a portable target to the Gruntfile (it kept building for Linux, instead!). However, I can produce a nice portable binary by `$(npm bin)/build --win portable` in the `electronuserland/builder:wine` Docker image. If this could be implemented properly, it would solve #397 and #649 properly.

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -93,9 +93,16 @@ if (config.DEVELOPMENT) {
   app.commandLine.appendSwitch('ignore-certificate-errors', 'true');
 }
 
-if (argv.portable) {
-  const EXEC_PATH = process.env.APPIMAGE || process.execPath;
-  const USER_PATH = path.join(EXEC_PATH, '..', 'Data');
+///////////////////////////////////////////////////////////////////////////////
+// Set data path for portable versions
+///////////////////////////////////////////////////////////////////////////////
+if (process.env.PORTABLE_EXECUTABLE_DIR || argv.portable) {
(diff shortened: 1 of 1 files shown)


sergeevabc:
Either Wire’s installer should have portable option (see [WikidPad][1]’s installation routine for example) which allows a user to choose the path or Wire should be deployed as ZIP package as well (see [Telegram][2]’s releases page for example). It would greatly ease the pain of installing and _updating_ (yes. it’s a huge pain at present) for those of us who use Wire on the go.

[1]: https://vgy.me/yS1Nkn.png
[2]: https://github.com/telegramdesktop/tdesktop/releases

ConorIA:
@sergeevabc, `electron-builder` provides the option to create an NSIS installer with a modifiable install path, as well as a zip file, however, only the `portable` target (which is a self-contained `.exe` provides the `PORTABLE_EXECUTABLE_DIR` environment variable, which obviates the need to pass the `--portable` flag; this, I expect, is desirable.

ConorIA:
I did some testing under a Windows 10 virtual machine and under Wine. In both instances, trying to set the user dir to `PORTABLE_EXECUTABLE_DIR` didn't work. Closing this as it didn't accomplish what I had thought it would.

sergeevabc:
:(

PR #1029, opened 2017-10-11
Make about selectable (Fixes #989)

### LINKED ISSUE #989: Wire's "About" dialog does not allow copying.
Oh boy, this is fun. Ran into a bug while reporting another bug.

Wire Version: 3.0.2814
Wire for Web Version: 2017.09.26.1300
OS: macOS Sierra 10.12.6
Do you have an antivirus software installed: Nope

> What steps will reproduce the problem?

1. Open the "About" dialog.
2. Try to copy text from there.
3. Break down in tears because it does not work.

> What is the expected result?

That you can select copy the text displayed there.

> What happens instead?

It won't let you select anything, thus preventing you from copying any text. This is pretty toxic, considering you want this information for your bug reports. And other apps usually allow this... see Finder, VLC Player, VirtualBox, pretty much any app I tried.

lipis:


electron/css/about.css
```diff
@@ -31,11 +32,9 @@ body {
   display: flex;
   justify-content: center;
   align-items: center;
-  font-family: -apple-system, ".SFNSText-Regular", "Helvetica Neue", "Segoe UI", Arial, sans-serif;
+  font-family: -apple-system, '.SFNSText-Regular', 'Helvetica Neue', 'Segoe UI', Arial, sans-serif;
   font-size: 11px;
   text-align: center;
-  user-select: none;
```
gregor: Should we make everything user selectable or just the version information?

lipis: Don't see any special reason to have only version if we are about to make it selectable.. :)

We could also simply copy that info with Ctrl/Cmd+C in that window.. like any prompt dialog boxes on Windows!! (That thing I miss on macOS)

lipis:
So shall we make only the versions copyable or is good to go?

gregor:
version

Let's make only these two lines copyable.

PR #1035, opened 2017-10-12
Prettier :)

lipis:
#1058 should pass first..


lipis:
Will do it from scratch.. it's way too old!

PR #1036, opened 2017-10-12
fix: WEBAPP-4392, WEBAPP-4396

pwnsdx:


gregor:
Proposed fix for WEBAPP-4396 is still buggy. Env does not prevail over selected env in internal. See #1037

pwnsdx:
That's another issue, my fix allowed me to select "Production" and it worked well, anyway I'm closing this ticket.

PR #1038, opened 2017-10-12
fix: WEBAPP-4392

pwnsdx:


tests/certPinningTest.js
```diff
@@ -0,0 +1,85 @@
...
+    );
+
+    Promise.all(promises)
+      .then(hostnames => {
+        hostnames.forEach(data => assert(certutils.verifyPinning('app.wire.com', data)));
+        done()
+      });
+  });
+
+  it('doesn\'t pin other hostnames', () => {
```
lipis: avoid escaping with "

ffflorian: If you say so, I will do that. :stuck_out_tongue:

tests/certPinningTest.js
```diff
@@ -0,0 +1,86 @@
...
+const https = require('https');
+
+const assert = require('assert');
+const path = require('path');
+
+const buildCert = cert => `-----BEGIN CERTIFICATE-----\n${cert.raw.toString('base64')}\n-----END CERTIFICATE-----`;
+
+const goodURLs = [
+  'app.wire.com',
+  //'d3o4r5p7qlbt6o.cloudfront.net',
```
gregor: Let's remove this and then it is good to go with the addition of tests.

PR #1045, opened 2017-10-16
fix: Sidebar context menu behaviour (WEBAPP-4312)

Yserz:


electron/renderer/src/actions/index.js
```diff
@@ -45,7 +45,21 @@ export const deleteAccount = id => {
   };
 };
 
-export const abortAccountCreation = (id) => {
+export const toggleAccountContextVisibility = (x, y) => {
```
gregor: Let's keep this in alphabetical order please

electron/renderer/src/components/Sidebar.js
```diff
@@ -66,24 +80,51 @@ const Sidebar = ({
...
+          forceVisible={isContextMenuVisible}
+        />
+      )}
 
     <ContextMenu id="account">
-      <ContextMenuItem onClick={() => window.open('https://wire.com/create-team/?pk_campaign=client&pk_kwd=desktop')}>
+      <ContextMenuItem
+        onClick={() =>
+          window.open(
+            'https://wire.com/create-team/?pk_campaign=client&pk_kwd=desktop'
```
gregor: We need to make this URL environment aware if we have the time to. Code style of these lines also looks off.

electron/renderer/src/components/context/ContextMenuTrigger.js
```diff
@@ -0,0 +1,40 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import React from 'react';
+
+const ContextMenuTrigger = ({ onClick, forceVisible }) => (
```
gregor: I would rename this as it is not a generic context menu trigger but the specific one for the add account/create team action.

electron/renderer/src/components/Sidebar.js
AndyTheGiant: Let's do this in another PR.

electron/renderer/src/components/Sidebar.js
```diff
@@ -66,24 +80,51 @@ const Sidebar = ({
...
+  </div>
+);
 
-export default Sidebar;
+export default connect(
+  ({ accounts, contextMenuState }) => ({
+    accounts,
+    currentAccentID: (accounts.find(account => account.visible) || {}).accentID,
+    hasCreatedAccount: accounts.some(account => account.userID !== undefined),
+    hasReachedLimitOfAccounts: accounts.length === 3,
```
gregor: Let's be safer in this case and make sure that if someone managed to create more than three account - however that might be possible (changing the content of the database, etc) - the test still works. => `accounts.length >= 3`

PR #1054, opened 2017-10-17
app.exit() instead of return

lipis:
Not sure how can we test it.. but for sure `return` was doing nothing in that case.

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -118,7 +118,7 @@ if (environment.platform.IS_WINDOWS) {
 
   // Stop further execution on update to prevent second tray icon
   if (shouldQuit) {
-    return;
+    app.exit();
   }
 }
 


gregor:
But nothing is exactly what it was supposed to do here. Why do you want to change it?

lipis:
Because this `return` has no effect inside the `if`.. so we can basically remove these lines.. if not exiting the app

gregor:
?!?

I think you got confused here for a moment. It is an `if` statement. 

````
function test() {
  if (1 === 1) {
    return;
  }
  console.log('I got here');
}
````

Will never print the `console.log`

lipis:
this will never print because you have a function.. but in our case there is no function: 

https://github.com/wireapp/wire-desktop/blob/master/electron/main.js#L113-L123 (I was talking about the `if` in the line 113)

lipis:
#1058 you can see the error there :)

gregor:
The return does exactly what it is supposed to be doing. I can show you at the office tomorrow. Have a safe flight.

lipis:
Oki!! 🇩🇰 ✈️ 🇩🇪

lipis:
After some tests.. https://github.com/eslint/eslint/issues/8767 I found this..

PR #1059, opened 2017-10-17
fix: Compile libsodium-neon with correct x86 linker and module version on Windows

mythsunwind:
Most linker issues in the rust compilation were fixed by setting the Visual Studio development environment with VsDevCmd.bat.

The remaining conflict between the node module machine type and the target machine type was solved by using the x86 version of node.


gregor:
Amazing 🎆

PR #1089, opened 2017-11-05
Create a better .desktop file

ConorIA:
### Rationale

- `.desktop` spec is updated to version 1.1
- adds keywords and generic name
- uses standard category names
- uses "Wire" as application menu name, not `wire-desktop`

### Result: 

```
[Desktop Entry]
Name=Wire
Comment=Secure messenger for everyone.
Exec="/opt/wire-desktop/wire-desktop" %U
Terminal=false
Type=Application
Icon=wire-desktop
Version=1.1
GenericName=Privacy-Oriented Messenger
Keywords=chat;encrypt;e2e;messenger;videocall
StartupWMClass=Wire
Categories=Network;InstantMessaging;Chat;VideoConference;
```

Edit: I would like developer input re: [keywords, GenericName](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s05.html), and [Categories](https://specifications.freedesktop.org/menu-spec/latest/apa.html#main-category-registry) [+](https://specifications.freedesktop.org/menu-spec/latest/apas02.html).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,6 +25,14 @@ const ELECTRON_PACKAGE_JSON = 'electron/package.json';
 const PACKAGE_JSON = 'package.json';
 const INFO_JSON = 'info.json';
 
+const DESKTOP_SPEC = {
+  "Version": "1.1",
+  "Name": "Wire",
+  "GenericName": "Privacy-Oriented Messenger",
+  "Keywords": "chat;encrypt;e2e;messenger;videocall",
+  "StartupWMClass": "Wire"
+};
+
 module.exports = function(grunt) {
   require('load-grunt-tasks')(grunt, {pattern: ['grunt-*']});
   const path = require('path');
@@ -146,10 +154,8 @@ module.exports = function(grunt) {
             executableName: 'wire-desktop',
             afterInstall: 'bin/deb/after-install.tpl',
             afterRemove: 'bin/deb/after-remove.tpl',
-            desktop: {
-              "StartupWMClass": "Wire"
-            },
-            category: 'Network',
+            desktop: DESKTOP_SPEC,
+            category: 'Network;InstantMessaging;Chat;VideoConference',
             depends: ['libappindicator1', 'libasound2', 'libgconf-2-4', 'libnotify-bin', 'libnss3', 'libxss1'],
           },
         },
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -25,6 +25,14 @@ const ELECTRON_PACKAGE_JSON = 'electron/package.json';
 const PACKAGE_JSON = 'package.json';
 const INFO_JSON = 'info.json';
 
+const DESKTOP_SPEC = {
+  "Version": "1.1",
+  "Name": "Wire",
+  "GenericName": "Privacy-Oriented Messenger",
```
lipis: `Secure messenger`

Gruntfile.js
```diff
@@ -25,6 +25,14 @@ const ELECTRON_PACKAGE_JSON = 'electron/package.json';
 const PACKAGE_JSON = 'package.json';
 const INFO_JSON = 'info.json';
 
+const DESKTOP_SPEC = {
+  "Version": "1.1",
```
lipis: What version is that and where it appears?

lipis: Maybe we can just use the `'<%= info.version %>'` (as in other places).

Also use single quotes for strings.. `'`

Gruntfile.js
```diff
@@ -146,10 +154,8 @@ module.exports = function(grunt) {
             executableName: 'wire-desktop',
             afterInstall: 'bin/deb/after-install.tpl',
             afterRemove: 'bin/deb/after-remove.tpl',
-            desktop: {
-              "StartupWMClass": "Wire"
-            },
-            category: 'Network',
+            desktop: DESKTOP_SPEC,
+            category: 'Network;InstantMessaging;Chat;VideoConference',
```
lipis: maybe we could put that in a variable as well and rename the other one to something like: `LINUX_DESKTOP` and `LINUX_CATEGORY`.

WDYT?

Gruntfile.js
mythsunwind: It is the version of the desktop entry specification: https://standards.freedesktop.org/desktop-entry-spec/1.1/apfs02.html

lipis: Right.. :)

lipis: Also mentioned in the description... I'm blind :D

Gruntfile.js
ConorIA: Is the category here used anywhere other than the `.desktop` file? It works fine if I just nest the categories in the desktop spec.

ConorIA:
😄 thanks @lipis and @mythsunwind!

PR #1101, opened 2017-11-10
fix: Hardening “About Wire” window (WEBAPP-4478)

pwnsdx:
- Now using Electron sandbox, preload script, ipc, nodeIntegration on false and javascript is also disabled on rendered page
- Prevent dropping links / following any kind of links but
- Only allow https:// links to be opened in the user browser (Chrome, Edge, FF…)
- Prevent the window from being minimized/maximized
- Always put the window on top of the other windows
- Typos


electron/main.js
```diff
@@ -24,12 +24,17 @@ const fs = require('fs-extra');
 const minimist = require('minimist');
 const path = require('path');
 const raygun = require('raygun');
-const { app, BrowserWindow, ipcMain, Menu, shell } = require('electron');
+const { session, app, BrowserWindow, ipcMain, Menu, shell } = require('electron');
```
lipis: no bracket spacing and put the contents in alphabetical order

electron/main.js
```diff
@@ -298,23 +304,83 @@ const showMainWindow = () => {
...
       height: config.WINDOW.ABOUT.HEIGHT,
+      maximizable: false,
+      minimizable: false,
       resizable: false,
-      fullscreen: false
+      show: false,
+      title: config.NAME,
+      width: config.WINDOW.ABOUT.WIDTH,
+      backgroundColor: '#ececec',
+      webPreferences: {
```
lipis: sort all properties in alphabetical order

electron/main.js
```diff
@@ -298,23 +304,83 @@ const showMainWindow = () => {
...
+        session: session.fromPartition('about-window'),
+        javascript: false,
+        nodeIntegration: false,
+        nodeIntegrationInWorker: false,
+        preload: path.join(APP_PATH, 'js', 'about.js'),
```
gregor: If this is now a preload file, this needs to be reflected in the name.

yarn.lock
```diff
@@ -0,0 +1,6619 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
gregor: No yarn on wire-desktop

.gitignore
```diff
@@ -14,3 +14,4 @@ app.asar
 node_modules
 npm-debug.log
 Wire*.zip
+yarn.lock
```
gregor: Alphabetical order if you really want to add it here.

PR #1125, opened 2017-11-28
[WEBAPP-4491] fix: Strip mailto: from copied links

ffflorian:
In addition to https://github.com/wireapp/wire-webapp/pull/1957.


electron/js/menu/context.js
```diff
@@ -119,7 +119,7 @@ window.addEventListener('contextmenu', (event) => {
     imageMenu.popup(remote.getCurrentWindow());
   } else if (element.nodeName === 'A') {
     event.preventDefault();
-    copyContext = element.href;
+    copyContext = element.href.replace(/^mailto:/, '');
```
bennyn: Can't we just do `substr`?

PR #1149, opened 2017-12-05
style: Change color of webview close button (WEBAPP-4524)

gregor:


electron/renderer/src/components/Webviews.css
```diff
@@ -29,6 +29,6 @@
   position: absolute;
   top: 24px;
   right: 24px;
-  fill: #fff;
+  fill: #33373a;
```
lipis: my favorite color :D

PR #1178, opened 2017-12-22
fix: close button should not be visible after login (WEBAPP-4562)

AndyLnd:


electron/renderer/src/components/Webviews.js
```diff
@@ -63,7 +70,7 @@ class Webviews extends Component {
   }
 
   _onUnreadCountUpdated(accountId, unreadCount) {
-    this.props.updateAccountBadgeCount(account.id, unreadCount);
+    this.props.updateAccountBadgeCount(accountId, unreadCount);
```
AndyLnd: I am not sure if and why this even worked ...

gregor: It did not. Was broken since https://github.com/wireapp/wire-desktop/pull/1162

PR #1186, opened 2018-01-01
Pop!_OS-indicator

AtomicCons:
Added support for Pop!_OS indicator.
Requires user to have appindicator extenstion installed - extension install is not required prior to wire install


AtomicCons:
pulled the wrong change.

PR #1253, opened 2018-02-05
WEBAPP-4478: Fix whitelist issue on Windows 

pwnsdx:


gregor:
This PR does not move but only add the dependency to the electron package json in this form. Will take care of it.

PR #1339, opened 2018-03-13
fix: prevent panel from opening when dragging the window

AndyLnd:


electron/js/preload.js
```diff
@@ -65,6 +65,24 @@ const addDragRegion = () => {
...
     titleBar.className = 'drag-region';
     document.body.appendChild(titleBar);
+
+    let isDragging = false;
+    let isMoved = false;
+
+    titleBar.addEventListener('mousedown', () => (isDragging = true));
+    document.addEventListener('mousemove', () => (isMoved = isDragging));
+
+    titleBar.addEventListener('mouseup', e => {
```
gregor: Characters are free. We still don't like non descriptive variable names even if it is `e`

AndyLnd:
Handled here now: https://github.com/wireapp/wire-webapp/pull/2798

PR #1347, opened 2018-03-20
chore: Remove libsodium-neon

ffflorian:
Since libsodium with WASM support [is included in the webapp](https://github.com/wireapp/wire-webapp/pull/2565), we can remove libsodium-neon and the Rust requirement.


bennyn:
I only approve this with a laughing and a crying eye... 😭 🔮

ffflorian:
tick :cry:

maximbaz:
Huh, so we can remove `cargo` from the dependencies of AUR packages all together now...

Just curious, do you have some info on perf comparison between wasm and rust solutions?

ffflorian:
Hi @maximbaz,

yes - we compared the functions [`@wireapp/proteus`](https://github.com/wireapp/wire-web-packages/tree/master/packages/proteus) uses and this is the result:

comparison

(`crypto_auth_hmacsha256_verify` was not measured because it was not implemented in `libsodium-neon`).

Alltogether we loose a tiny bit of speed (you won't even notice) but we get rid of a lot of building errors, platform specific instructions and the big Rust dependency :slightly_smiling_face:

PR #1484, opened 2018-05-24
Added Catalan locale

bgta:
I pushed Catalan locale for Desktop app.


gregor:
@bgta Thanks for your effort. Translations are handled via crowdin though as described in our readme (https://github.com/wireapp/wire-desktop#translations). We will add the locale once both webapp and desktop app are at 100%. I will therefore have to close this PR.

PR #1607, opened 2018-07-13
 feat: remove team creation from wrapper (WEBAPP-5164)

AndyLnd:
This is related to https://github.com/wireapp/wire-webapp/pull/4038


electron/renderer/src/components/Webviews.js
```diff
@@ -59,13 +59,17 @@ class Webviews extends Component {
...
-  _getEnvironmentUrl(account) {
+  _getEnvironmentUrl(account, forceLogin) {
     const envParam = decodeURIComponent(new URL(window.location).searchParams.get('env'));
     const url = new URL(envParam);
 
     // pass account id to webview so we can access it in the preload script
     url.searchParams.set('id', account.id);
 
+    if (forceLogin) {
+      url.hash = '#login';
```
gregor: Let's try to not go this path again. The dependency between browser and webapp should be as limited as possible. We already have outdated wrapper versions that contained this at some point 🙈

PR #1618, opened 2018-07-18
fix: Aligns delete action copies between webapp and wrapper

atomrc:


gregor:
LGTM. Let's keep in mind to start the localization process in time before the release.

PR #1631, opened 2018-07-23
fix incorrect spelling of dependency for rpm

silvernode:
Package installation fails on Fedora 28:
- nothing provides Gconf2 needed by wire-desktop-3.2.0_511072-1.i386

Correct package name:
- GConf2


silvernode:
Yay! Glad I could help I will keep the commit title structure in mind for next time.

gregor:
@silvernode No worries. We use semantic commit messages in order to aid us when generating change logs.

PR #1634, opened 2018-07-23
chore: do not ignore yarn.lock to detect wrong usage of yarn on the project

atomrc:


ffflorian:
I thought we couldn't use `yarn` on this project because it doesn't support lifecycle scripts? Is that resolved?

PR #1645, opened 2018-07-27
fix: Prevent download and parsing of non html files (WEBAPP-4422)

atomrc:


electron/js/lib/openGraph.js
```diff
@@ -41,8 +41,21 @@ const fetchImageAsBase64 = url => {
...
 const fetchOpenGraphData = url => {
-  return new Promise((resolve, reject) => {
-    openGraph(url, (error, meta) => (error ? reject(error) : resolve(meta)));
+  const isHtmlContentTypePromise = new Promise(resolve => {
+    const headerRequest = request.head(url);
+    headerRequest.on('response', response => {
+      if (response.headers['content-type'] && response.headers['content-type'].match(/.*text\/html/)) {
+        resolve(true);
+      }
+      resolve(false);
```
gregor: We can simplify this. Move it to a const that is either true or false and the resolve that.

atomrc: :man_facepalming:

PR #1671, opened 2018-08-14
fix: Limit content length size when downloading link preview (WEBAPP-4422, WEBAPP-5341)

atomrc:
Previous strategy was:

- do HEAD request and read the content of the response ;
- if the content type doesn't match throw error ;
- else send another request via `openGraph`.

New strategy:

- do full request ;
- when the headers arrive (before any content is downloaded), check that `content-type` is of type `text/html` ;
- start counting downloaded bytes ;
- if the number of bytes exceed the limit, we fail ;
- if not we call the `parse` method of `openGraph` only on the `<head>` of the document.

### DIFF
--- a/electron/js/lib/openGraph.js
+++ b/electron/js/lib/openGraph.js
@@ -47,11 +47,7 @@ const fetchOpenGraphData = url => {
   const normalizedUrl = parsedUrl.protocol ? parsedUrl : urlUtil.parse(`http://${url}`);
 
   const parseHead = body => {
-    const headMatches = body.match(/<head>[\s\S]*?<\/head>/);
-    if (!headMatches) {
-      throw new Error('No head found in the document');
-    }
-    const [head] = headMatches;
+    const [head] = body.match(/<head>[\s\S]*?<\/head>/) || [''];
     return openGraphParse(head);
   };
 
--- a/electron/js/lib/openGraph.js
+++ b/electron/js/lib/openGraph.js
@@ -52,7 +52,7 @@ const fetchImageAsBase64 = url => {
 };
 
 const fetchOpenGraphData = url => {
-  const CONTENT_SIZE_LIMIT = 1000000; // ~1MB
+  const CONTENT_SIZE_LIMIT = 1e6; // ~1MB
   const parsedUrl = urlUtil.parse(url);
   const normalizedUrl = parsedUrl.protocol ? parsedUrl : urlUtil.parse(`http://${url}`);
 
--- a/electron/js/lib/openGraph.js
+++ b/electron/js/lib/openGraph.js
@@ -29,12 +29,22 @@ const bufferToBase64 = (buffer, mimeType) => {
 };
 
 const fetchImageAsBase64 = url => {
+  const IMAGE_SIZE_LIMIT = 5e6; // 5MB
   return new Promise(resolve => {
-    request({encoding: null, url: encodeURI(url)}, (error, response, body) => {
+    const imageRequest = request({encoding: null, url: encodeURI(url)}, (error, response, body) => {
       if (!error && response.statusCode === 200) {
         resolve(bufferToBase64(body, response.headers['content-type']));
       } else {
         // we just skip images that failed to download
+        resolve();
+      }
+    });
+
+    let currentDownloadSize = 0;
+    imageRequest.on('data', buffer => {
+      currentDownloadSize += buffer.length;
+      if (currentDownloadSize > IMAGE_SIZE_LIMIT) {
+        imageRequest.abort();
         resolve();
       }
     });
--- a/electron/js/lib/openGraph.js
+++ b/electron/js/lib/openGraph.js
@@ -42,7 +42,7 @@ const fetchImageAsBase64 = url => {
 };
 
 const fetchOpenGraphData = url => {
-  const CONTENT_TYPE_LIMIT = 1e7; // 10MB
+  const CONTENT_TYPE_LIMIT = 1000000; // 1MB
   const parsedUrl = urlUtil.parse(url);
   const normalizedUrl = parsedUrl.protocol ? parsedUrl : urlUtil.parse(`http://${url}`);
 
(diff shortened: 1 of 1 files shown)


electron/js/lib/openGraph.js
```diff
@@ -41,21 +42,37 @@ const fetchImageAsBase64 = url => {
...
+        throw new Error(`Unhandled format for open graph generation ('${contentType}' of size '${contentLength}')`);
+        getContentRequest.abort();
+      }
+    });
+
+    let requestCurrentSize = 0;
+    getContentRequest.on('data', buffer => {
+      requestCurrentSize += buffer.length;
+      if (requestCurrentSize > CONTENT_TYPE_LIMIT) {
+        throw new Error(`File size too big for open graph generation ('${contentLength}')`);
```
pwnsdx: Error throwing must come after aborting the request, they act as a return.

electron/js/lib/openGraph.js
```diff
@@ -41,21 +42,37 @@ const fetchImageAsBase64 = url => {
...
-  });
+    getContentRequest.on('response', ({headers}) => {
+      const contentType = headers['content-type'] || '';
+      const contentLength = parseInt(headers['content-length'] || '0', 10);
+
+      const isHtmlContentType = contentType.match(/.*text\/html/);
+      const isTooLarge = contentLength > CONTENT_TYPE_LIMIT;
+
+      if (!isHtmlContentType || isTooLarge) {
+        throw new Error(`Unhandled format for open graph generation ('${contentType}' of size '${contentLength}')`);
```
pwnsdx: Error throwing must come after aborting the request, they act as a return.

electron/js/lib/openGraph.js
gregor: Let's try to not through but parse partial downloads. Open graph data can be found in `<meta>` tags within then `<head>`. Thus we can set the limit much, much lower and even check for the closing head tag.

electron/js/lib/openGraph.js
```diff
@@ -41,21 +42,37 @@ const fetchImageAsBase64 = url => {
...
-      return new Promise((resolve, reject) => openGraph(url, (error, meta) => (error ? reject(error) : resolve(meta))));
-    }
-    throw new Error('Unhandled format for open graph generation');
-  });
+    getContentRequest.on('response', ({headers}) => {
+      const contentType = headers['content-type'] || '';
+      const contentLength = parseInt(headers['content-length'] || '0', 10);
+
+      const isHtmlContentType = contentType.match(/.*text\/html/);
+      const isTooLarge = contentLength > CONTENT_TYPE_LIMIT;
```
pwnsdx: If you really want to check the content type and its length then you need to restore the HEAD request otherwise those checks only happen at the end.

electron/js/lib/openGraph.js
```diff
@@ -41,21 +42,37 @@ const fetchImageAsBase64 = url => {
...
 const fetchOpenGraphData = url => {
-  const isHtmlContentTypePromise = new Promise(resolve => {
-    const headerRequest = request.head(url);
-    headerRequest.on('response', response => {
-      const isHtmlContentType =
-        response.headers['content-type'] && response.headers['content-type'].match(/.*text\/html/);
-      resolve(isHtmlContentType);
+  const CONTENT_TYPE_LIMIT = 1e7; // 10MB
+  const parsedUrl = urlUtil.parse(url);
+  const normalizedUrl = parsedUrl.protocol ? parsedUrl : urlUtil.parse(`http://${url}`);
```
pwnsdx: Request should not happen if no protocol are set.

pwnsdx:
Downloading images should also be subject to size limitation: https://github.com/wireapp/wire-desktop/pull/1671/files#diff-5d44f5d58de8cfcdc007eebe371758a9R33

Unifying the two functions in one would be ideal.

electron/js/lib/openGraph.js
atomrc: The strategy changed with my last commit

electron/js/lib/openGraph.js
atomrc: That's actually a fix for a bug we have. That's expected behavior

electron/js/lib/openGraph.js
atomrc: The `on('response')` event is actually called before any content is downloaded (basically when the headers are here)

electron/js/lib/openGraph.js
```diff
@@ -41,21 +42,38 @@ const fetchImageAsBase64 = url => {
...
+        throw new Error(`Unhandled format for open graph generation ('${contentType}')`);
+      }
+    });
+
+    let requestCurrentSize = 0;
+    let partialBody = '';
+    getContentRequest.on('data', buffer => {
+      const chunk = buffer.toString();
+      requestCurrentSize += buffer.length;
+      partialBody += chunk;
```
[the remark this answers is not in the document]
atomrc: Humm I would argue that this is not the same thing. `buffer.length` is the number of bytes while `body.length` is the number of characters (that could differ because of unicode).

electron/js/lib/openGraph.js
pwnsdx: Headers are part of the downloaded content. I think you need a HEAD request.

pwnsdx: Another approach if we don't want to do 2 requests would be extension blacklisting but it's far from perfect.

electron/js/lib/openGraph.js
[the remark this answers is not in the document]
pwnsdx: Exact but it does not need to be that precise ;)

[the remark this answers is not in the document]
pwnsdx: (If you feel you need to keep this precision then I would just suggest to rename `requestCurrentSize` to `responseCurrentSize`)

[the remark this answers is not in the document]
atomrc: The number of characters is good enough indeed

electron/js/lib/openGraph.js
atomrc: I checked, this function is called before any `data` event occurs. We are pretty safe on that side :+1:

atomrc: You can find the lyfecycle of a request here https://nodejs.org/api/http.html#http_http_request_options_callback

Actually if the body is empty, no `data` event will be emitted.

electron/js/lib/openGraph.js
```diff
@@ -28,34 +29,63 @@ const bufferToBase64 = (buffer, mimeType) => {
 };
 
 const fetchImageAsBase64 = url => {
+  const IMAGE_SIZE_LIMIT = 5e6; // 5MB
```
lipis: 👍

electron/js/lib/openGraph.js
```diff
@@ -28,34 +29,63 @@ const bufferToBase64 = (buffer, mimeType) => {
...
 };
 
 const fetchOpenGraphData = url => {
-  const isHtmlContentTypePromise = new Promise(resolve => {
-    const headerRequest = request.head(url);
-    headerRequest.on('response', response => {
-      const isHtmlContentType =
-        response.headers['content-type'] && response.headers['content-type'].match(/.*text\/html/);
-      resolve(isHtmlContentType);
+  const CONTENT_SIZE_LIMIT = 1000000; // ~1MB
```
lipis: `1e6`?

electron/js/lib/openGraph.js
```diff
@@ -28,34 +29,63 @@ const bufferToBase64 = (buffer, mimeType) => {
 };
 
 const fetchImageAsBase64 = url => {
+  const IMAGE_SIZE_LIMIT = 5e6; // 5MB
   return new Promise(resolve => {
-    request({encoding: null, url: encodeURI(url)}, (error, response, body) => {
+    const imageRequest = request({encoding: null, url: encodeURI(url)}, (error, response, body) => {
       if (!error && response.statusCode === 200) {
```
lipis: can we start avoiding magic numbers?

maybe adding all the status codes in one file and have it like `status.OK`?

Could be part of another PR though..

electron/js/lib/openGraph.js
atomrc: absolutely :)

electron/js/lib/openGraph.js
atomrc: Good point but, yeah, probably not the scope of this PR

PR #1675, opened 2018-08-15
fix: JavaScript errors popping up when updating tray icon badge (WEBAPP-4558)

bennyn:
## What happened until now
- Our Electron app is distributed as an [asar package](https://github.com/electron/asar), which is an extensive archive format similar to `tar`
- When executing our Electron app, the contents of the asar package get automatically unarchived to the `%LocalAppData%` directory on Windows
- Third-party performance tools (like [CCleaner](https://www.ccleaner.com/)) clean up files from `%LocalAppData%` during runtime and mistakenly remove unarchived resources from our app
- When receiving a new message, while Wire is running in the background, we want to signal this by showing a badge on the Wire tray icon; thus we need to load a different image which will then be used for the tray icon
- The image which we want to load might not be present anymore, because CCleaner deleted it already, which will cause [TypeError: Error processing argument at index 0](https://github.com/electron/electron/issues/7657#issuecomment-255673246) when passing a path of an invalid file to `appIcon.setImage` or `win.setOverlayIcon`

## Solution
- On app start all extracted resources are present (because CCleaner can only delete them after they have been created), so we can initially read our tray icon images from the disk
- Instead of reading the images from disk everytime we want to do a tray icon image update, we will cache them on the initial app start using Electron's [nativeImage API](https://github.com/electron/electron/blob/v2.0.7/docs/api/native-image.md#nativeimage) which loads images into memory
- Every visual tray icon update will then be made using the cached instance of the image we want to display
- When the extracted resources get deleted, this won't be a problem for us because we still have the cached instance and on a new app start all resources get extracted again, so we can recreate the image cache

## Benefits
- Total robustness to deleting files from `%LocalAppData%\wire\app-[version]\resources\app.asar\img`
- Less I/O operations because we can load icons from memory instead of accessing them from disk on every icon image update
- Only small footprint on RAM (our tray icon images on Windows have a total size of 58.8 KB)

## Implementation
- Instead of exporting single functions to do updates on the tray icon, there is now a `TrayIconHandler` which manages the tray icon state and its updates
- The `TrayIconHandler` automatically creates a [Tray](https://electronjs.org/docs/api/tray) on instantiation but it also accepts existing tray icons and mocked tray icons (for testing purposes)
- Platform-specific features have been outsourced to getter attributes like `hasTrayMenuSupport` & `hasOverlaySupport`
- The `updateBadgeIcon` test now runs on all platforms and supports asynchronous behaviour

## Where to go from here
- Currently we use `*.ico` files [for Windows](https://github.com/electron/electron/blob/master/docs/api/native-image.md#supported-formats) and `*.png` files for all other platforms; we can consider moving to a PNG-only approach since [high-resolution PNG images](https://electronjs.org/docs/api/native-image#high-resolution-image) are supported with Electron
- When we need to cache alot more resources, then we can see what it means to move from an asar packaging to shipping our app by copying all of its source files


electron/js/menu/TrayIconHandler.js
```diff
@@ -0,0 +1,85 @@
...
+        label: locale.getText('trayQuit'),
+      },
+    ]);
+
+    this.appIcon.on('click', () => windowManager.showPrimaryWindow());
+    this.appIcon.setContextMenu(contextMenu);
+    this.appIcon.setToolTip(config.NAME);
+  }
+
+  createIcons() {
```
gregor: No need to expose this outside of the constructor

electron/js/menu/TrayIconHandler.js
```diff
@@ -0,0 +1,85 @@
...
+      ),
+    };
+  }
+
+  updateBadgeIcon(win, count) {
+    if (count) {
+      this.appIcon.setImage(this.icons.trayWithBadge);
+    } else {
+      this.appIcon.setImage(this.icons.tray);
+    }
```
gregor:
````
const image = count ? this.icons.trayWithBadge : this.icons.tray;
this.appIcon.setImage(image);
````

File: electron/js/menu/TrayIconHandler.js [hunk 3]
```diff
@@ -0,0 +1,85 @@
...
+
+    if (this.hasOverlaySupport) {
+      win.setOverlayIcon(count ? this.icons.badge : null, locale.getText('unreadMessages'));
```
gregor:
````
const overlayImage = count ? this.icons.badge : null;
win.setOverlayIcon(overlayImage , locale.getText('unreadMessages'));
````

File: electron/js/menu/TrayIconHandler.js [hunk 4]
```diff
@@ -0,0 +1,85 @@
...
+
+class TrayIconHandler {
+  constructor(appIcon = new Tray(nativeImage.createEmpty())) {
+    this.appIcon = appIcon;
+    this.icons = this.createIcons();
+    this.lastUnreadCount = 0;
+    this.initTrayIcon();
+  }
+
+  initTrayIcon() {
```
gregor: No need to expose this to the public

electron/js/menu/TrayIconHandler.js
```diff
@@ -0,0 +1,85 @@
...
+
+  get hasTrayMenuSupport() {
+    return !environment.platform.IS_MAC_OS;
+  }
+
+  get hasOverlaySupport() {
+    return environment.platform.IS_WINDOWS;
+  }
+
+  buildTrayMenu() {
```
gregor: No need to expose this to the public

electron/js/menu/TrayIconHandler.js
atomrc: Because we don't really have a way to create private method and we might not want to be fancy and play with `call`, `apply` or `bind` a solution could be just to inline this function in the constructor.
It's a pretty small function, so I think it's fine to inline

electron/js/menu/TrayIconHandler.js
atomrc: The case of this one is a bit trickier because it's a bit bigger. 
So, in order not to bloat the public API of this class, a few solutions include:

- still inlining it in the `constructor` ;
- define it as a external function and call it with `call` (`buildTrayMenu.call(this)`) ;
- use the `_` convention to prevent misuse of the method. 

I think I would for the the `call` solution but it's up for discussion of course.

electron/js/menu/TrayIconHandler.js
atomrc: This one doesn't need `this` at all, so it could definitely be defined in the private scope

electron/js/menu/TrayIconHandler.js
```diff
@@ -0,0 +1,85 @@
...
+      this.appIcon.setImage(this.icons.trayWithBadge);
+    } else {
+      this.appIcon.setImage(this.icons.tray);
+    }
+
+    if (this.hasOverlaySupport) {
+      win.setOverlayIcon(count ? this.icons.badge : null, locale.getText('unreadMessages'));
+    }
+
+    if (win.isFocused() || !count) {
```
atomrc: how about `win.flashFrame(!win.isFocused() && count && count > this.lastUnreadCount)`

(the condition could be extracted in a variable)

gregor: I think this condition is becoming very complicated that way. We had a lot of issues with this and @maximbaz fixed it before.

maximbaz: Also I believe it won't be equal, it is crucial that we don't call flashFrame in some cases

atomrc: Not sure I get your point guys? `flashFrame` is called either way (falsy or truthy cases)

If you find the condition to hard to read, it could be decomposed into two 

```
const hasFocus = win.isFocused();
const hasNewMessages = count && count > this.lastUnreadCount;

win.flashFrame(hasFocus && hasNewMessages)
```

atomrc: Ohhh ok, I re-read the `else if`, indeed there are some cases when the method is not called.
Since I don't know the consequences of that method, maybe it's better to leave it like this then

maximbaz: If the window is not focused, count is 2 and lastUnreadCount is 5, we dont want to call flashFrame at all

bennyn: @maximbaz Thanks for posting this scenario. I will create a test case from it. 😎

tests/main/TrayIconHandlerTest.js
```diff
@@ -0,0 +1,79 @@
...
+        useContentSize: true,
+      });
+
+      sinon.spy(appWindow, 'flashFrame');
+
+      appWindow.loadURL(`file://${path.join(__dirname, '..', 'fixtures', 'badge.html')}`);
+      appWindow.webContents.on('dom-ready', () => {
+        assert.equal(appWindow.isFocused(), false);
+        tray.updateBadgeIcon(appWindow, 2);
+        assert.equal(appWindow.flashFrame.getCall(0), null);
```
atomrc: :+1:

electron/js/menu/TrayIconHandler.js
```diff
@@ -0,0 +1,81 @@
...
+      tray: nativeImage.createFromPath(path.join(app.getAppPath(), 'img', `tray.${this.defaultImageExtension}`)),
+      trayWithBadge: nativeImage.createFromPath(
+        path.join(app.getAppPath(), 'img', `tray.badge.${this.defaultImageExtension}`)
+      ),
+    };
+
+    this.initTrayIcon();
+  }
+
+  initTrayIcon() {
```
atomrc: I would still inline this method in the `constructor` so that the API of the class is pristine from methods that should not be used outside :)

bennyn: I accept payments via PayPal or Bitcoin. 😁

tests/main/TrayIconHandlerTest.js
```diff
@@ -21,10 +21,13 @@
 
 const assert = require('assert');
 const path = require('path');
-const sinon = require('sinon');
+const chai = require('chai');
```
atomrc: out of curiosity, what is the benefit of `chai` over `sinon`?

electron/js/menu/TrayIconHandler.js
```diff
@@ -0,0 +1,77 @@
...
+  constructor(platform, appIcon = new Tray(nativeImage.createEmpty())) {
+    this.appIcon = appIcon;
+    this.lastUnreadCount = 0;
+    this.platform = platform;
+
+    this.icons = {
+      badge: nativeImage.createFromPath(path.join(app.getAppPath(), 'img', 'taskbar.overlay.png')),
+      tray: nativeImage.createFromPath(path.join(app.getAppPath(), 'img', `tray.${this.defaultImageExtension}`)),
+      trayWithBadge: nativeImage.createFromPath(
+        path.join(app.getAppPath(), 'img', `tray.badge.${this.defaultImageExtension}`)
```
gregor: Do we want to pull this out into a const for improved readability? Will be two instead of three lines as well.

tests/main/TrayIconHandlerTest.js
```diff
@@ -0,0 +1,89 @@
...
+describe('TrayIconHandler', () => {
+  const TrayMock = {
+    on: () => {},
+    setContextMenu: () => {},
+    setImage: () => {},
+    setToolTip: () => {},
+  };
+
+  describe('"constructor"', () => {
+    it('creates native images for all tray icons on instantiation.', () => {
```
gregor: Are we changing the signature of our tests naming? We never ended them with a .

PR #1676, opened 2018-08-16
chore: add tray icon resolutions

AndyLnd:


electron/js/menu/TrayHandler.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+class TrayHandler {
+  constructor() {
+    this.hasOverlaySupport = false;
+    this.lastUnreadCount = 0;
+  }
+
+  initTray(hasOverlaySupport, trayIcon = new Tray(nativeImage.createEmpty())) {
```
atomrc: Why moving this part to another method? This creates *temporal coupling* (the consumer now has to call the methods in the right order else he will have unexpected behaviors)

bennyn:
## Changes
1. Before this PR we used `*.ico` files for Windows and `*.png` files for all other platforms. Now only `*.png` is being used (also on Windows).
1. Tray icons are now available in 11 different resolutions to [support high-DPI screens](https://github.com/electron/electron/blob/master/docs/api/native-image.md#high-resolution-image).
1. Tray icons are now organized in separate directories.
1. Since all platforms are using `*.png` images, we have less platform-specific differences in handling the tray icon. The only difference now is that [Windows supports taskbar overlays](https://electronjs.org/docs/api/browser-window#winsetoverlayiconoverlay-description-windows). That's why the dependency on `platform` could be completely removed and is now replaced by a simple `hasOverlaySupport` flag.
1. `updateBadgeIcon` has been renamed to `showUnreadCount` because it does more than just updating the icon. That's why `showUnreadCount` now internally calls three different methods.

electron/js/menu/TrayHandler.js
bennyn: I want the consumer to decide if he/she wants to have a tray icon or not. That's why it got decoupled.

atomrc: Very valid point!

radarlog:
icons scaling probably does not work in Cinnamon (Manjaro Linux). The default `img/tray-icon/tray/tray.png` icon is using at HiDPI resolutions

PR #1677, opened 2018-08-16
fix: Don't show tray icon on Mac

bennyn:


electron/js/menu/TrayIconHandler.js
```diff
@@ -52,15 +52,14 @@ class TrayIconHandler {
...
-  get hasTrayMenuSupport() {
-    return !this.platform.IS_MAC_OS;
-  }
-
   get hasOverlaySupport() {
     return this.platform.IS_WINDOWS;
   }
 
   updateBadgeIcon(win, count) {
+    if (!this.appIcon) {
```
gregor: This is not specific to Windows. The `flashFrame` and `setBadgeCount` also need to be called on macOS

PR #1702, opened 2018-09-06
fix: Keep app position and size on macOS when exiting (WEBAPP-5423)

bennyn:


bennyn:
Closed in favor of https://github.com/wireapp/wire-desktop/pull/1706

PR #1706, opened 2018-09-07
fix: Keep app position and size on macOS when exiting (WEBAPP-5423)

bennyn:


bennyn:
TBD: Implement test with
- no `window-state.json`
- a `init.json` with `fullscreen: true` but without `bounds`

electron/main.js
```diff
@@ -136,16 +162,19 @@ const showMainWindow = () => {
...
-  const windowBounds = settings.restore(SETTINGS_TYPE.WINDOW_BOUNDS, main.getBounds());
-  main.setBounds(windowBounds);
-  if (isFullScreen) {
+  main = new BrowserWindow(options);
+  mainWindowState.manage(main);
+  if (typeof mainWindowState.isMaximized === 'undefined' && isFullScreen === true) {
     main.maximize();
   }
 
+  upgradeSettingsToV1();
```
bennyn: This should only be done if there is an `init.json` file and if that file does not contain a `configVersion` property.

electron/js/lib/settings.js
```diff
@@ -21,15 +21,27 @@
...
 const path = require('path');
-
-const app = require('electron').app || require('electron').remote.app;
 const debug = require('debug');
+const mkdirp = require('mkdirp');
+const app = require('electron').app || require('electron').remote.app;
 
-const INIT_JSON = path.join(app.getPath('userData'), 'init.json');
+const configDir = path.join(app.getPath('userData'), 'config');
+mkdirp.sync(configDir);
```
bennyn: Instead of a combination of `fs` and `mkdirp`, we can simply use `fs-extra` for both tasks (creating files and creating directories).

electron/js/lib/upgrade-settings-v1.js
```diff
@@ -0,0 +1,47 @@
...
+const mkdirp = require('mkdirp');
+
+const debugLogger = debug('UpgradeInitFile');
+const settings = require('./settings');
+const SETTINGS_TYPE = require('./settingsType');
+const configDir = path.join(app.getPath('userData'), 'config');
+
+mkdirp.sync(configDir);
+
+const upgradeSettingsToV1 = () => {
```
bennyn: Moving the config files can be also part of this upgrade process.

electron/js/lib/settings.js
ffflorian: But `mkdirp` was already in the dependencies.

electron/js/lib/upgrade-settings-v1.js
bennyn: `fs-extra` too :)

electron/main.js
```diff
@@ -116,14 +118,37 @@ const bindIpcEvents = () => {
...
+    isFullScreen = showInFullScreen;
+  }
+
+  return WindowStateKeeper(stateKeeperOptions);
+};
+
 // App Windows
-const showMainWindow = () => {
-  const showMenuBar = settings.restore(SETTINGS_TYPE.SHOW_MENU_BAR, true);
+const showMainWindow = mainWindowState => {
```
atomrc: Big :+1:  to those injected config params

electron/js/menu/system.js
```diff
@@ -390,11 +390,11 @@ const changeLocale = language => {
 };
 
 module.exports = {
-  createMenu: () => {
+  createMenu: isFullScreen => {
     if (environment.platform.IS_MAC_OS) {
       menuTemplate.unshift(darwinTemplate);
       windowTemplate.submenu.push(separatorTemplate, showWireTemplate, separatorTemplate, toggleFullScreenTemplate);
-      toggleFullScreenTemplate.checked = settings.restore(SETTINGS_TYPE.FULL_SCREEN, false);
+      toggleFullScreenTemplate.checked = isFullScreen || false;
```
gregor: Weird condition on an or. Let's put this as the default in the param instead as we are overwriting a false with a false. And we are actually doing it twice.

electron/js/menu/system.js
```diff
@@ -410,7 +410,7 @@ module.exports = {
         i18n: 'menuPreferences',
       });
       windowTemplate.submenu.push(separatorTemplate, toggleMenuTemplate, separatorTemplate, toggleFullScreenTemplate);
-      toggleFullScreenTemplate.checked = settings.restore(SETTINGS_TYPE.FULL_SCREEN, false);
+      toggleFullScreenTemplate.checked = isFullScreen || false;
```
gregor: See above.

electron/main.js
```diff
@@ -68,6 +69,7 @@ let tray = undefined;
 
 let main;
 let isQuitting = false;
+let isFullScreen = false;
```
gregor: ABC

electron/main.js
```diff
@@ -116,14 +118,37 @@ const bindIpcEvents = () => {
   ipcMain.on(EVENT_TYPE.WRAPPER.RELAUNCH, lifecycle.relaunch);
 };
 
+const initWindowStateKeeper = () => {
+  const loadedWindowBounds = settings.restore(SettingsType.WINDOW_BOUNDS, {
+    height: config.WINDOW.MAIN.DEFAULT_HEIGHT,
+    width: config.WINDOW.MAIN.DEFAULT_WIDTH,
+  });
+
+  const showInFullScreen = settings.restore(SettingsType.FULL_SCREEN, undefined);
```
gregor: Let's not set the default to undefined. This is either true or false. Maybe we should also move the getters for the individual settings to the ConfigurationPersistence and collect the defaults there instead of having them spread all over the app.

electron/main.js
```diff
@@ -136,13 +161,14 @@ const showMainWindow = () => {
...
+    y: mainWindowState.y,
+  };
 
-  const isFullScreen = settings.restore(SETTINGS_TYPE.FULL_SCREEN, false);
-  const windowBounds = settings.restore(SETTINGS_TYPE.WINDOW_BOUNDS, main.getBounds());
-  main.setBounds(windowBounds);
-  if (isFullScreen) {
+  main = new BrowserWindow(options);
+  mainWindowState.manage(main);
+  if (typeof mainWindowState.isMaximized === 'undefined' && isFullScreen === true) {
```
gregor: Why undefined? This should be true or false. Also: Does this account for the difference between macOS and Windows?

ffflorian: This is part of the upgrade process to `electron-window-state`. This line comes only to use if there is no `isMaximized` value (which comes from electron-window-state) and if we still have the old setting `isFullScreen` set to `true`. In the future the can remove this line as `electron-window-state` will maximize the window for us if appropriate.

gregor: Let's write the code in a way that this is clear to everyone then. The line above does not hint at being related to the upgrade process at all. Pull it into a descriptive const and add a comment.

electron/main.js
ffflorian: This is again the case of the old config file. I am now setting it explicitly to 'not-set-in-v0' instead of undefined.

PR #1715, opened 2018-09-12
fix: Load the full (clear|set)Immediate polyfill needed by the webapp

atomrc:
This is the second step of https://github.com/wireapp/wire-webapp/pull/4552

### DIFF
--- a/electron/renderer/static/webview-preload.js
+++ b/electron/renderer/static/webview-preload.js
@@ -191,8 +191,10 @@ const checkAvailability = callback => {
 
 // https://github.com/electron/electron/issues/2984
 const _setImmediate = setImmediate;
+const _clearImmediate = clearImmediate;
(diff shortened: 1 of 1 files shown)


electron/renderer/static/webview-preload.js
```diff
@@ -191,8 +191,10 @@ const checkAvailability = callback => {
 
 // https://github.com/electron/electron/issues/2984
 const _setImmediate = setImmediate;
+const _clearImmediate = clearImmediate;
 process.once('loaded', () => {
   global.setImmediate = _setImmediate;
+  global.clearImmediate = _clearImmediate;
   global.desktopCapturer = desktopCapturer;
   global.environment = environment;
   global.openGraph = require('../../js/lib/openGraph');
```
gregor: Let's do ABC for readability

PR #1724, opened 2018-09-13
refactor: Rename files containing JSX from .js to .jsx

Yserz:


gregor:
@Yserz With releases on all platforms out this is good to go whenever you are ready.

Issue #1075, opened 2017-10-24
--portable flag useless on Linux when installing from deb


linluc:
Wire Version: 3.0.2816
Wire for Web Version: 2017.10.19.1045
OS: Linux Mint 18.1, 64bit
Do you have an antivirus software installed: No

What steps will reproduce the problem?
1. Download wire deb package. Here "wire_3.0.2816_amd64.deb" was used.
2. Install Wire from the deb package using either system built-in tools or manually via dpkg in terminal.
3. Run "wire-desktop --portable" as regular user

What is the expected result?

Wire starts up normally.  --portable flag should only change the location of the "data" directory.

What happens instead?

Nothing happens, Wire does not start.

Please provide any additional information below. Attach a screenshot if
possible.

Executing `strace -e trace=file wire-desktop --portable` shows the following lines at the end:

stat("/opt/wire-desktop/Data", 0x7ffc86051b40) = -1 ENOENT (No such file or directory)
mkdir("/opt/wire-desktop/Data", 0700)   = -1 EACCES (Permission denied)
stat("/opt/wire-desktop/Data", 0x7ffc86051b40) = -1 ENOENT (No such file or directory)
readlink("/opt/wire-desktop/Data/SS", 0x7ffc860505d0, 4096) = -1 ENOENT (No such file or directory)
readlink("/opt/wire-desktop/Data/SingletonLock", 0x7ffc860503d0, 4096) = -1 ENOENT (No such file or directory)
symlink("mint-twin1-4122", "/opt/wire-desktop/Data/SingletonLock") = -1 ENOENT (No such file or directory)
readlink("/opt/wire-desktop/Data/SingletonLock", 0x7ffc8604fff0, 4096) = -1 ENOENT (No such file or directory)
readlink("/opt/wire-desktop/Data/SS", 0x7ffc860505d0, 4096) = -1 ENOENT (No such file or directory)
readlink("/opt/wire-desktop/Data/SingletonLock", 0x7ffc860503d0, 4096) = -1 ENOENT (No such file or directory)
+++ exited with 0 +++

The reason of that behavior is that the directory where Wire got installed i.e. "/opt/wire-desktop" can only be written to by root. 3 possible solutions:

1. Allow Wire install in any user provided directory. Currently it seems impossible with a deb package as /opt/wire-desktop is hardcoded. At least I was unable to overwrite it with dpkg switches.

[...]

PR #1730, opened 2018-09-14
feat: Add option to set user data path

gregor:
#1078 related to #1075 has been hanging around way too long. Sorry about that 🙈

@ConorIA Since our codebase has evolved and your PR has merge conflicts I simply created a new one. I opted for a naming similar to the start param of Google Chrome (https://chromium.googlesource.com/chromium/src/+/HEAD/docs/user_data_dir.md).

Unless there are strong opinions agains this I would want to merged this and thereby fix and address #1075 and #1078.

### DIFF
--- a/electron/js/appInit.js
+++ b/electron/js/appInit.js
@@ -37,10 +37,14 @@ const fixUnityIcon = () => {
   }
 };
 
-const handlePortableFlag = () => {
-  if (argv.portable) {
+const handlePortableFlags = () => {
+  if (argv.portable || argv.user_data_dir) {
+    const directory = argv.user_data_dir ? argv.user_data_dir : 'Data';
+
     const EXEC_PATH = process.env.APPIMAGE || process.execPath;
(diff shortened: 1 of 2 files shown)


electron/js/appInit.js
```diff
@@ -37,10 +37,14 @@ const fixUnityIcon = () => {
...
 
-const handlePortableFlag = () => {
-  if (argv.portable) {
+const handlePortableFlags = () => {
+  if (argv.portable || argv.user_data_dir) {
+    const directory = argv.user_data_dir ? argv.user_data_dir : 'Data';
+
     const EXEC_PATH = process.env.APPIMAGE || process.execPath;
-    const USER_PATH = path.join(EXEC_PATH, '..', 'Data');
+    const USER_PATH = path.join(EXEC_PATH, '..', directory);
```
atomrc: I think it is not going to work if an absolute path are given by the user.
It's going to generate currentDir/../<start of absolute path>

PR #1803, opened 2018-10-15
feat: Add mute shortcut (WEBAPP-5587)

atomrc:


gregor:
Yes. This menu option is no longer supposed to be there.

PR #1805, opened 2018-10-16
fix: Check certificate verification result

ffflorian:
This PR will
* log the hostname when a SSL connection failed
* use the verification result for validating SSL connections


gregor:
As discussed f2f: zinfra.io cannot be used as internal environment

PR #1845, opened 2018-10-31
fix: Electron 3.0.6

pwnsdx:
- Fixed account switching shortcut
- Fixed environment swapping (via saving settings before leaving the app)
- Fixed WEBAPP-5587
- Refactored App.jsx


electron/src/js/menu/system.ts
```diff
@@ -399,6 +399,20 @@ const createMenu = (isFullScreen: boolean): Menu => {
     helpTemplate.submenu = [];
   }
 
+  // Global mute shortcut
+  globalShortcut.register('CmdOrCtrl+Alt+M', () => sendAction(EVENT_TYPE.CONVERSATION.TOGGLE_MUTE));
+
+  // Global account switching shortcut
+  const switchAccountShortcut = ['CmdOrCtrl', 'Super'];
+  const accountLimit = 3;
```
bennyn: Is the account limit specified somewhere else? I could imagine we have it as a config setting already?

lipis: is that a hard limit?

package.json
```diff
@@ -58,7 +58,7 @@
   },
   "license": "LicenseRef-LICENSE",
   "lint-staged": {
-    "*.{js,jsx}": [
+    "*.{js}": [
```
bennyn: I will fix the JSX styling issue.

lipis: @pwnsdx you can use the `--no-verify` if that is causing some errors..

electron/src/js/menu/system.ts
bennyn: yes, it is

PR #1868, opened 2018-11-06
fix: Remove Google Contacts import (WEBAPP-5660)

pwnsdx:


.eslintignore
```diff
@@ -1,4 +0,0 @@
-node_modules
-electron/dist
-electron/node_modules
-electron/renderer/dist
```
lipis: Are you sure this is not needed?!

Try running the tests.. if it takes 3 hours, revert this file.

electron/src/js/lib/SingleSignOn.ts
```diff
@@ -0,0 +1,318 @@
...
+      if (this.session) {
+        this.session.clearStorageData({}, () => resolve());
+      } else {
+        resolve();
+      }
+    });
+  };
+}
+
+export {SingleSignOn};
```
lipis: I guess this should be part of a different PR.. as this is for removing the google import.

PR #1892, opened 2018-11-14
fix: Wrapper intercepts shortcuts when in background (WEBAPP-5704)

pwnsdx:


electron/src/main.ts
```diff
@@ -207,6 +207,9 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
     setTimeout(() => main.show(), 800);
   }
 
+  main.on('focus', () => systemMenu.createShortcuts());
```
bennyn: Let's call it `registerShortcuts` and `unregisterShortcuts`.

AndyLnd: I believe we also don't need to create new functions here and can just use
```main.on('focus', systemMenu.registerShortcuts);```

File: electron/src/main.ts [hunk 2]
```diff
@@ -207,6 +207,9 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
     setTimeout(() => main.show(), 800);
   }
 
+  main.on('focus', () => systemMenu.registerShortcuts());
```
AndyLnd: I believe we don't need to create new functions here. 
```main.on('focus', systemMenu.registerShortcuts);```
should do.

PR #1901, opened 2018-11-15
refactor: Move directories out of "js" subfolder

bennyn:


electron/src/lib/openGraph.ts
```diff
@@ -21,7 +21,7 @@ import * as request from 'request';
 import * as urlUtil from 'url';
 const {parse: openGraphParse} = require('open-graph');
 
-import {OpenGraphResult} from '../../interfaces/';
```
ffflorian: Why is this needed?

electron/src/lib/openGraph.ts
```diff
@@ -21,7 +21,7 @@ import * as request from 'request';
...
+import {OpenGraphResult} from '../interfaces/index';
```
ffflorian: Why is `index` needed here?

PR #1905, opened 2018-11-16
fix: Enable asar packaging on Linux

ffflorian:
Follow-up to https://github.com/wireapp/wire-desktop/pull/335.

Tested this on Debian with `en_US` locale.

@lipis @maximbaz @ConorIA
Can you test this, too? :slightly_smiling_face:

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -214,7 +214,7 @@ module.exports = function(grunt) {
 
       options: {
         arch: 'all',
-        asar: false,
+        asarUnpack: 'node_modules/spellchecker/vendor',
         publish: null,
       },
     },


maximbaz:
Oh this would be disappointing, just curious why do you want to enable it?

ffflorian:
@maximbaz 
Why disappointing? Didn't @ConorIA said that it worked for him, too?
The reason is that asar can considerably improve the startup time of the app.

maximbaz:
I said disappointing just because I make use of disabled asar to [enable dark theme](https://gist.github.com/maximbaz/e80a105f4f16813f4b2549e70787e984) for my instance of Wire, I know at least one other person who uses my theme as well 😄 

But I don't wanna block your PR, I can always disable asar locally when compiling 😉 

image

ffflorian:
@maximbaz Oh, cool :slightly_smiling_face: 
We could think about unpacking more files to also make @ConorIA happy (see https://github.com/wireapp/wire-desktop/issues/147).

maximbaz:
That would be nice, if you choose to unpack more files, for dark theme I would need access to `resources/app/main.js` and `resources/app/css/` dir 🙂 

I'm curious how much startup speed improvement do you see and whether this improvement is significantly affected if we start unpacking more files...

ffflorian:
> I'm curious how much startup speed improvement do you see and whether this improvement is significantly affected if we start unpacking more files...

Good question. Needs some more testing I guess :slightly_smiling_face:

ffflorian:
We will leave it like it is to not break people's build pipelines.

pwnsdx:
> That would be nice, if you choose to unpack more files, for dark theme I would need access to `resources/app/main.js` and `resources/app/css/` dir 🙂
> 
> I'm curious how much startup speed improvement do you see and whether this improvement is significantly affected if we start unpacking more files...

Have you tried [this option](https://electronjs.org/docs/tutorial/application-packaging#adding-unpacked-files-to-asar-archives), @maximbaz?

I think would be a great idea to reenable asar packing on Linux as soon as [Dark Mode](https://github.com/wireapp/wire-webapp/commit/5a3f5a6c647506ea4c1d960c8afc088bb8751c7d) is in production for the same reason as @ffflorian but also to align with the other platform.

maximbaz:
I haven't tried that option, but if I need to do something custom locally, I can just as well disable asar altogether 🙂 It seems we don't have any proof on it being a performance boost, other than that it is brings more annoyance than good.

By the way, kudos on implementing official dark mode @pwnsdx, I wanted to try but it seems README.md is outdated and `?env=prod` trick no longer works on `dev` branch. Can't wait for it to reach production!

PR #1914, opened 2018-11-19
fix: Use correct event channel, use correct devTools

ffflorian:


electron/src/menu/developer.ts
```diff
@@ -37,7 +37,7 @@ const devToolsTemplate: Electron.MenuItemConstructorOptions = {
   submenu: [
     {
       accelerator: 'Alt+CmdOrCtrl+I',
-      click: () => (getPrimaryWindow() as any).toggleDevTools(),
+      click: () => getPrimaryWindow().webContents.toggleDevTools(),
```
ffflorian: Has the same effect, but this is the correct way to do it.

electron/src/menu/system.ts
```diff
@@ -74,7 +75,7 @@ const localeTemplate: ElectronMenuItemWithI18n = {
 };
 
 const aboutTemplate: ElectronMenuItemWithI18n = {
-  click: (): void => (menu as any).emit(EVENT_TYPE.ABOUT.SHOW),
+  click: () => ipcMain.emit(EVENT_TYPE.ABOUT.SHOW),
```
ffflorian: We shouldn't send custom events over the Menu's IPC channel.

PR #1915, opened 2018-11-19
fix: Undo/Redo behavior

pwnsdx:


lipis:
Maybe adding tests?

bennyn:
@lipis We will run some tests on RC builds and then merge the PR if tests are passed.

PR #1920, opened 2018-11-20
fix: Crash on main process when log does not exist

bennyn:


electron/src/util/LogFactory.ts
```diff
@@ -50,10 +50,14 @@ class LogFactory {
...
-        encoding: 'utf8',
-        flag: 'a',
-      });
+      try {
+        fs.writeFileSync(LogFactory.getFileURI(), `${withoutColor}\r\n`, {
+          encoding: 'utf8',
+          flag: 'a',
+        });
+      } catch (error) {
+        // Don't crash on I/O logging errors
```
ffflorian: Let's maybe add a console log at least?

bennyn: ok

PR #1924, opened 2018-11-21
refactor: Do not show window title for SSO window when the origin is the backend

pwnsdx:


electron/src/lib/SingleSignOn.ts
```diff
@@ -170,6 +171,10 @@ class SingleSignOn {
   // Ensure the requested URL is going to the backend
   public static isBackendOrigin = (url: string) => SingleSignOn.ALLOWED_BACKEND_ORIGINS.includes(new URL(url).origin);
 
+  // Get window title
```
bennyn:
```suggestion
```

PR #1929, opened 2018-11-23
fix: Don't create link previews for not found pages (WEBAPP-5730)

ffflorian:


electron/src/lib/openGraph.ts
```diff
@@ -64,12 +64,15 @@ const fetchOpenGraphData = (url: string): Promise<OpenGraphResult> => {
   };
 
   return new Promise<string>((resolve, reject) => {
-    const getContentRequest = request.get(urlUtil.format(normalizedUrl), (error, response, body: string) => {
-      return error ? reject(error) : resolve(body);
```
ffflorian: This was basically always resolving since we had `.on('data')` with a `resolve` and this callback.

electron/src/lib/openGraph.ts
```diff
@@ -58,18 +58,16 @@ const fetchOpenGraphData = (url: string): Promise<OpenGraphResult> => {
...
-
   return new Promise<string>((resolve, reject) => {
-    const getContentRequest = request.get(urlUtil.format(normalizedUrl), (error, response, body: string) => {
-      return error ? reject(error) : resolve(body);
-    });
+    let partialBody = '';
+    const getContentRequest = request.get(urlUtil.format(normalizedUrl));
+
+    getContentRequest.on('response', response => {
+      if (response.statusCode !== 200) {
```
ffflorian: How should we handle redirects?

atomrc: I don't think we should trigger an error for redirects.
If you link `website.com` and you are redirected to a valid `www.website.com` you would not be please if the link preview doesn't show ... I guess.

ffflorian: Seems like the `statusCode` is checked after the redirect.

PR #1943, opened 2018-11-27
fix: Ensure folder is created when writing into the log file

pwnsdx:


ffflorian:
@pwnsdx @bennyn can we delete this branch?

Issue #1944, opened 2018-11-28
Provide PGP signature for sources tarball as part of a release


maximbaz:
I've learned a lot about how package managers work on Linux and what are the usual ways of verifying authenticity of downloaded source code when compiling a package. I'd like to share what I know and ask for your help on improving the distribution, consider this a follow-up to #542. cc @raphaelrobert 

1. Please upload [your PGP key](https://wire-app.wire.com/linux/releases.key) to a key server such as http://keys.gnupg.net/.

    Package managers like `pacman` are integrated with `gpg`, which in turn knows how to automatically fetch keys from a key server. Package maintainers will only have to specify your PGP key ID, and package manager will do the rest: fetch your key from a key server and use it to validate the downloaded sources.

    There are many keyservers, but you only have to upload your key to one of them, they are periodically synchronizing between themselves.

1. Please provide a `.sig` file for the sources tarball as part of a release.

    When you create a release, Github automatically attaches a source tarball `3.4.2879.tar.gz`:

    ![image](https://user-images.githubusercontent.com/1177900/49120574-7ae6a500-f2ad-11e8-9064-b8c9b5a8c76f.png)

    This is the sources archive that package maintainers will use. When a `.sig` file is present (`3.4.2879.tar.gz.sig` in this example), package managers like `pacman` that are integrated with `gpg` will automatically use it to verify the signature of the downloaded sources. It's pretty neat as allows to avoid having to trust package maintainer that they used valid sources when packaging the app.

    Here's how you would create a signature:

    ```
    # First generate the tarball that Github will generate when creating the release
    # This is guaranteed to generate bit-by-bit identical archive
    git archive -o 3.4.2879.tar.gz --format tar.gz --prefix=wire-desktop-release-3.4.2879/ release/3.4.2879

    # Now generate a PGP signature for this file, this creates `3.4.2879.tar.gz.sig` that you need to upload as part of a release
    gpg --detach-sign --yes --local-user D599C1AA126762B1 3.4.2879.tar.gz

    # Finally remove the tarball
    rm 3.4.2879.tar.gz
    ```

------

cc @ConorIA, after this is implemented we could apply the patch below and pacman will do the rest for us.

```patch
--- PKGBUILD    2018-11-28 01:35:16.940349157 +0100
+++ PKGBUILD-new        2018-11-28 01:36:36.186535102 +0100
@@ -16,9 +16,12 @@
             'ttf-emojione: for colorful emoji made by EmojiOne')
[...]

PR #1947, opened 2018-11-28
chore: Sign .tar.gz sources with Jenkins

ffflorian:
This fixes https://github.com/wireapp/wire-desktop/issues/1944.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,3 +15,21 @@
 *.ttf binary
 *.woff binary
 *.woff2 binary
+
+# Ignore files when creating archive
+.DS_Store export-ignore
+.grunt export-ignore
+.Icon export-ignore
+.vscode export-ignore
+*.log export-ignore
+*.pyc export-ignore
+/.idea export-ignore
+/electron/dist export-ignore
+/electron/node_modules export-ignore
+/electron/renderer/dist export-ignore
+/keys export-ignore
+/temp export-ignore
+/wrap export-ignore
+app.asar export-ignore
+node_modules export-ignore
+Wire*.zip export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -24,7 +24,6 @@
 *.log export-ignore
 *.pyc export-ignore
 *.pkg export-ignore
-info.json export-ignore
 /.idea export-ignore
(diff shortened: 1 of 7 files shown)


.gitattributes
```diff
@@ -15,3 +15,21 @@
...
+/.idea export-ignore
+/electron/dist export-ignore
+/electron/node_modules export-ignore
+/electron/renderer/dist export-ignore
+/keys export-ignore
+/temp export-ignore
+/wrap export-ignore
+app.asar export-ignore
+node_modules export-ignore
+Wire*.zip export-ignore
```
ffflorian: Taken from `.gitignore`.

bin/linux-checksums.sh
```diff
@@ -51,23 +68,30 @@ if ! ls ./*.AppImage > /dev/null 2>&1; then
...
 
-_log "Create checksums..."
+_log "Creating checksums..."
 sha256sum *.deb *.rpm *.AppImage > sha256sum.txt
 
-_log "Prepare gpg configuration..."
-mkdir -p "${GPG_TEMP_DIR2}"
+_log "Creating source code archive for signing..."
+(
+  cd ../../
```
ffflorian: We are in `wrap/dist` but need to be in the main folder of the repo.

bin/linux-checksums.sh
```diff
@@ -1,15 +1,34 @@
...
-SCRIPT_DIR="${0%/*}"
 
 GPG_TEMP_DIR=".gpg-temporary"
-GPG_TEMP_DIR2=".gpg-temporary/private-keys-v1.d"
+GPG_TEMP_KEYS_DIR="${GPG_TEMP_DIR}/private-keys-v1.d"
 PGP_SIGN_ID="D599C1AA126762B1"
 PGP_KEYFILE="${PGP_PRIVATE_KEY_FILE:-${PGP_SIGN_ID}.asc}"
 PGP_PASSPHRASE="${PGP_PASSPHRASE:-""}"
+SHRED_STATUS="available"
+BUILD_VERSION="${1}"
```
mythsunwind: Maybe it is good to move this on top to see that this is a needed parameter for the script

Issue #1949, opened 2018-11-28
Wire icon in system tray / notification are stretched when not using display scaling


Raytar:
Wire version: 3.4.2885 Wire for web version: 2018.11.15.1314 Operating system: Windows 10 1809 and Ubuntu 18.10 Which antivirus software do you have installed: Windows Defender

What steps will reproduce the problem?

1. Launch wire
2. Observe the icon in the notification area / system tray

What is the expected result? The icon should be round.
What is the actual result? The icon is stretched.

![image](https://user-images.githubusercontent.com/10762082/49148463-bca43980-f307-11e8-9fd2-8dbe7eaa6194.png)

If I set display scaling to 125% it looks fine.
The file [wire-desktop/electron/img/tray-icon/tray/tray.png](https://github.com/wireapp/wire-desktop/blob/a901d53d7e30433b4287f658ccc59b09654dfeb5/electron/img/tray-icon/tray/tray.png) seems to be the culprit as it is not square.

PR #1951, opened 2018-11-28
fix: Tray icon on Linux and Windows

ffflorian:
It's easier to just use the "old" (small) icon as default and the "new" (large) icon on Linux.

This fixes https://github.com/wireapp/wire-desktop/issues/1949


electron/src/js/environment.ts
```diff
@@ -86,6 +86,12 @@ const platform = {
   IS_WINDOWS: process.platform === 'win32',
 };
 
+const linuxDesktop = {
+  isGnome: (process.env.XDG_CURRENT_DESKTOP || '').includes('GNOME'),
```
AndyLnd: Please avoid the repetition here.

PR #1953, opened 2018-11-28
 fix: Force env variable to always be uppercase globally

pwnsdx:


electron/src/js/environment.ts
```diff
@@ -73,7 +73,7 @@ const app = {
 };
 
 const getEnvironment = (): TYPE => {
-  return currentEnvironment ? currentEnvironment : restoreEnvironment();
+  return <TYPE>(currentEnvironment ? currentEnvironment : restoreEnvironment()).toUpperCase();
```
ffflorian: Let's rename `TYPE` to something more related like `BackendLabelType`.

PR #1985, opened 2018-12-11
Remove 'Wire / Preferences' menu on Linux

votdev:
Because this is a duplicate of 'Edit / Preferences' which is the typical location for Linux apps.

Signed-off-by: Volker Theile <votdev@gmx.de>


ffflorian:
@votdev Thank you!

PR #1996, opened 2018-12-16
fixed: Remove unused global (WEBAPP-5814)

pwnsdx:


ffflorian:
Can we also delete the file then? I don't see it being used anywhere anymore.

PR #2018, opened 2018-12-20
refactor: Certificate verification improvements

pwnsdx:


electron/src/lib/CertificateVerifyProcManager.ts
```diff
@@ -40,6 +40,7 @@ interface DisplayCertificateErrorOptions {
 class CertificateVerifyProcManager {
   private static bypassCertificatePinning: boolean = false;
   private static isDialogLocked: boolean = false;
+  public static mainWindow: any;
```
ffflorian: Please add the correct type here.
```suggestion
  public static mainWindow: Electron.BrowserWindow;
```

PR #2044, opened 2018-12-28
Create a detached signature for the sources archive

maximbaz:
Follow-up to #1947 

Fixes https://github.com/wireapp/wire-desktop/issues/1944#issuecomment-449444326


ConorIA:
Seconded! :nerd_face:

Issue #2046, opened 2018-12-28
Page down and Page up keys disabled when Wire is in tray.


Tuj3Bz90dV:
Due to [this piece of code](https://github.com/wireapp/wire-desktop/blob/5a0a4fcdcbbb2faba616b60d1050be45b7df2e82/electron/src/menu/system.ts#L458-L460), the page up and down keys are globally overridden. This should probably not use the `globalShortcut` module.

This causes, in Linux at least, the page up and down keys to not work in other applications when Wire is minimized to tray.

Aranad:
I'm getting this issue to on Mint 19

ffflorian:
Hi, thanks for reporting.
This behavior is strange, since Electron should [unregister all shortcuts on blur](https://github.com/wireapp/wire-desktop/blob/master/electron/src/main.ts#L186) and on Debian / Gnome I can't reproduce that. How do you guys minimize the app? With a shortcut or by clicking somewhere?

Tuj3Bz90dV:
`Window -> Close`. The unregistering of shortcuts works when minimizing or simply losing focus. I suppose the question is whether the `hide()` event also triggers `blur` or not. I'm on XFCE, for whatever that's worth.

ffflorian:
@Tuj3Bz90dV
Ah, I see. I can reproduce that.

ffflorian:
This fix will be included in the next release.

terba:
I compiled Wire from the current sources some days ago, but the problem is still there. What can I do to solve this? I can not work without page up and down, so I have to quit Wire when this happens (many times per day).

ffflorian:
@terba when was "some days" ago? Try building from latest master.

terba:
It was compiled from good sources (on 2019-01-22), and it works well! I've played with the keyboard settings to solve the problem (before I found out that Wire causes the problem), and some apps were not working correctly because of this. Sorry for taking your time!

ffflorian:
Alright :slightly_smiling_face:

hylkevds:
For me this problem still exists, or maybe it exists again.

PR #2053, opened 2019-01-02
fix: Unregister global shortcuts on close event

ffflorian:
This fixes https://github.com/wireapp/wire-desktop/issues/2046.

### DIFF
--- a/electron/src/main.ts
+++ b/electron/src/main.ts
@@ -182,9 +182,6 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
     setTimeout(() => main.show(), 800);
   }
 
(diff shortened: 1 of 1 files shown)


electron/src/main.ts
```diff
@@ -223,6 +223,7 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
         main.hide();
       }
     }
+    systemMenu.unregisterShortcuts();
```
atomrc: Will it trigger a `focus` when the app is opened again? Else, it seems the shortcuts will never be registered again.

ffflorian: Yes, the `focus` event is triggered when clicking on the taskbar icon.

PR #2093, opened 2019-01-14
feat: Build using custom product name (WEBAPP-5875, WEBAPP-5877)

ffflorian:


deployment.Jenkinsfile
```diff
@@ -53,13 +54,17 @@ node('master') {
     currentBuild.displayName = "Deploy $projectName " + version
 
     stage('Upload to S3 and/or Hockey') {
-        if(projectName.contains('Windows')) {
+        if (projectName.contains("Windows")) {
```
lipis: inconsistent quotes.. let's use either double or single.. but not both :) (line 8 has single ones)

ffflorian: Usually we only use double quotes in Jenkinsfiles when variables are inserted, so I should revert the double quotes.

linux.Jenkinsfile
```diff
@@ -15,10 +15,11 @@ node('node180') {
   def text = readFile('info.json')
   def buildInfo = parseJson(text)
   def version = buildInfo.version + '.' + env.BUILD_NUMBER
-  currentBuild.displayName = version;
   def app_base = params.APP_BASE
 
-  def environment = docker.build("node", "-f linux.Dockerfile .")
+  currentBuild.displayName = version;
```
mythsunwind: no need for the semicolon here (the line was only moved, but i wanted to point this out ;) )

PR #2123, opened 2019-01-23
feat: Copy script with single files

ffflorian:
Follow-up to https://github.com/wireapp/wire-web-config-default/pull/7.

### DIFF
--- a/bin/configure.js
+++ b/bin/configure.js
@@ -18,11 +18,12 @@
  * along with this program. If not, see http://www.gnu.org/licenses/.
  *
  */
+
 /* eslint-disable no-console */
 
 const fs = require('fs-extra');
 const {execSync} = require('child_process');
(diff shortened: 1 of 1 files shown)


bin/configure.js
```diff
@@ -37,33 +38,52 @@ console.log(
...
 const configDirName = 'config';
 const configDir = resolve(configDirName);
-const src = resolve(configDir, pkg.name, 'content');
-const projectDir = resolve('.');
-const dest = projectDir;
-const ignoreList = ['.DS_Store'];
+const configSrc = resolve(configDir, pkg.name, 'content');
+const configDest = resolve('.');
+
+const files = [
```
bennyn: Let's turn it into an object like:

```js
const copy = {
  "image/logo/32x32.png": "resources/icons/32x32.png"
}
```

PR #2145, opened 2019-01-31
feat: wire:// custom protocol with "sso-code" module

pwnsdx:


electron/renderer/src/components/App.jsx
```diff
@@ -66,11 +91,13 @@ function mapStateToProps(state) {
   const {accounts} = state;
   return {
     accountIds: accounts.map(account => account.id),
+    accounts: accounts,
```
ffflorian:
```suggestion
    accounts,
```

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,121 @@
...
+      // App is booting
+      main.webContents.once('did-finish-load', () => sendCodeToRenderer());
+    } else {
+      if (!main.isVisible()) {
+        main.show();
+      }
+      sendCodeToRenderer();
+    }
+  };
+  private oldClipboard: string | undefined;
```
ffflorian:
```suggestion
  private oldClipboard?: string;
```

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,121 @@
...
+  private webview!: WebviewTag;
+  private webContents!: WebContents;
```
ffflorian: Can we do it without the `!`?

resources/macos/custom.plist
```diff
@@ -1,5 +1,16 @@
 <plist version="1.0">
 <dict>
+  <key>CFBundleURLTypes</key>
+  <array>
+      <dict>
+          <key>CFBundleURLName</key>
+          <string>Wire Core Protocol</string>
+          <key>CFBundleURLSchemes</key>
+          <array>
+            <string>wire</string>
```
ffflorian: This needs to be customizable in the Gruntfile

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: Readability vs actual gain of having the strict class initialization for those 2 vars is pretty bad, I think it's better to keep the !

electron/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "author": "Wire Swiss <wireapp@wire.com>",
+  "customProtocolName": "wire",
```
ffflorian: Better have the default in `CoreProtocol.ts`.
```suggestion
```

electron/src/lib/CoreProtocol.ts
```diff
@@ -0,0 +1,66 @@
...
+import {app} from 'electron';
+import * as path from 'path';
+import {URL} from 'url';
+import {AutomatedSingleSignOn} from './AutomatedSingleSignOn';
+
+const LOG_DIR = path.join(app.getPath('userData'), 'logs');
+const LOG_FILE = path.join(LOG_DIR, 'electron.log');
+const logger = LogFactory.getLogger('CoreProtocol', {forceEnable: true, logFilePath: LOG_FILE});
+
+const CORE_PROTOCOL = require('../../package.json').customProtocolName;
```
```suggestion
const {customProtocolName} = require('../../package.json');
const CORE_PROTOCOL = customProtocolName || 'wire';
```

info.json
```diff
@@ -3,6 +3,7 @@
   "appBase": "https://app.wire.com",
   "bundleId": "com.wearezeta.zclient.mac",
   "copyright": "© Wire Swiss GmbH",
+  "customProtocolName": "wire",
```
bennyn: Any change to reuse `process.env.APP_CUSTOM_PROTOCOL_NAME` here?

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,141 @@
...
+
+import {WebContents, WebviewTag, app, clipboard} from 'electron';
+import {BaseError} from 'make-error-cause';
+import {URL} from 'url';
+const dialog = require('electron').dialog || require('electron').remote.dialog;
+import * as windowManager from '../js/window-manager';
+import {getText} from '../locale/locale';
+import {EVENT_TYPE} from './eventType';
+
+class FailedToLoadWebviewError extends BaseError {}
```
bennyn: Why is there a need for `BaseError`?

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  public async restoreClipboard() {
+    if (this.oldClipboard) {
+      await AutomatedSingleSignOn.clipboard.write(this.oldClipboard);
+    }
+  }
+
+  public async start() {
+    this.oldClipboard = AutomatedSingleSignOn.clipboard.read();
+    await AutomatedSingleSignOn.clipboard.write(`${AutomatedSingleSignOn.SSO_CODE_PREFIX}${this.ssoCode}`);
```
bennyn: SSO code schema will be `wire://start-sso/wire-<UUID>`.

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    // Send initial signal to the renderer and wait for a response
+    window.addEventListener(EVENT_TYPE.ACTION.CREATE_ACCOUNT_RESPONSE, (event: any) => this.onResponseReceived(event), {
```
bennyn: Let's replace `any` with the actual type.

info.json
pwnsdx: Not inside the default info JSON file but we override here if `APP_CUSTOM_PROTOCOL_NAME` env is set: https://github.com/wireapp/wire-desktop/pull/2145/files#diff-35b4a816e0441e6a375cd925af50752cR55

bennyn: That's great! So we can have custom links for custom apps by relying on `baseData.customProtocolName`.

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+  private webContents!: WebContents;
+
+  constructor(private readonly ssoCode: string) {}
+
+  private buildUrl() {
+    if (!this.webview.src) {
+      return 'about:blank';
+    }
+    const src = new URL(this.webview.src);
+    src.pathname = '/auth/';
```
bennyn: `Webviews.jsx` also defines `url.pathname = '/auth'`. We should consider having a central place to define URLs.

Code from `Webviews.jsx`:

```jsx
  _getEnvironmentUrl(account, forceLogin) {
    const currentLocation = new URL(window.location.href);
    const envParam = currentLocation.searchParams.get('env');
    const decodedEnvParam = decodeURIComponent(envParam);
    const url = new URL(decodedEnvParam);

    // pass account id to webview so we can access it in the preload script
    url.searchParams.set('id', account.id);

    if (forceLogin) {
      url.pathname = '/auth';
      url.hash = '#login';
    }

    return url.href;
  }
```

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+    src.hash = AutomatedSingleSignOn.SSO_LOGIN_HASH;
+    return src.href;
+  }
+
+  private async clickOn(element: string) {
+    await this.webContents.executeJavaScript(`document.querySelector('${element}').click();`);
```
bennyn: Can we communicate with a `CustomEvent` rather than doing `executeJavaScript` to communicate between wrapper and webapp?

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+  private executeLogin() {
+    return new Promise((resolve, reject) => {
+      const failedToLoad = () => reject('Webview failed to load');
+      this.webview = document.querySelector('.Webview:not(.hide)') as WebviewTag;
+      this.webContents = this.webview.getWebContents();
+
+      this.webContents.once('did-fail-load', failedToLoad);
+      this.webContents.once('did-finish-load', async () => {
+        this.webContents.removeListener('did-fail-load', failedToLoad);
+        await this.clickOn('a[data-uie-name="go-sign-in-sso"]');
```
bennyn: Can we directly go to the link instead of going to a link via a simulated click?

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  public async restoreClipboard() {
+    if (this.oldClipboard) {
+      await AutomatedSingleSignOn.clipboard.write(this.oldClipboard);
+    }
+  }
+
+  public async start() {
+    if (this.ssoCode.length > AutomatedSingleSignOn.SSO_CODE_MAX_LENGTH) {
+      return;
```
bennyn: We should show an error in the view that the code is invalid rather than doing nothing.

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    this.oldClipboard = AutomatedSingleSignOn.clipboard.read();
+    await AutomatedSingleSignOn.clipboard.write(this.ssoCode);
```
bennyn: We can add a parameter that is read by the webapp to prefill the SSO login input.

electron/src/lib/CoreProtocol.ts
```diff
@@ -0,0 +1,80 @@
...
+import * as path from 'path';
+import {URL} from 'url';
+import {platform} from '../js/environment';
+import {AutomatedSingleSignOn} from './AutomatedSingleSignOn';
+
+const LOG_DIR = path.join(app.getPath('userData'), 'logs');
+const LOG_FILE = path.join(LOG_DIR, 'electron.log');
+const logger = LogFactory.getLogger('CoreProtocol', {forceEnable: true, logFilePath: LOG_FILE});
+
+const {customProtocolName} = require('../../package.json');
```
bennyn: `customProtocolName` seems to be defined in `info.json` and not `package.json`.

electron/src/lib/CoreProtocol.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  if (platform.IS_MAC_OS) {
+    app.on('open-url', async (event, url) => {
+      event.preventDefault();
+      await dispatcher(url);
+    });
+  } else if (platform.IS_WINDOWS) {
+    app.once('ready', async () => {
+      const url = process.argv[CORE_PROTOCOL_POSITION];
```
bennyn: Do we need this argument or is it for developer purposes only?

info.json
bennyn: We can convert `info.json` to `info.js` and replace `"wire"` with `process.env.APP_CUSTOM_PROTOCOL_NAME || 'wire'`.

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+    } catch (error) {
+      if (error instanceof MaximumAccountsReachedError) {
+        dialog.showMessageBox({
+          detail: getText('wrapperAddAccountErrorMessage'),
+          message: getText('wrapperAddAccountErrorTitle'),
+          type: 'warning',
+        });
+      }
+    } finally {
+      await this.restoreClipboard();
```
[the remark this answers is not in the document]
pwnsdx: We could override the navigator.clipboard.readText() function to return our value. Not sure it's the path we want to take considering this solution is very simple and the clipboard goes busy for a fraction of a second.

electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+    const src = new URL(this.webview.src);
+    src.pathname = '/auth/';
+    src.hash = AutomatedSingleSignOn.SSO_LOGIN_HASH;
+    return src.href;
+  }
+
+  private async clickOn(element: string) {
+    await this.webContents.executeJavaScript(`document.querySelector('${element}').click();`);
+    await AutomatedSingleSignOn.sleep(AutomatedSingleSignOn.SSO_ACTION_DELAY);
```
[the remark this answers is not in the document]
pwnsdx: We need to let time to the web app to react to simulated clicks.

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: We can't, Electron have an issue where the clipboard cannot be read *every time* if we directly go load the #sso page.

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: This is not an attempt to validate the code, rather to prevent pasting too big data from being pasted to the clipboard.

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: That would require changes on the web app.

electron/src/lib/CoreProtocol.ts
pwnsdx: It is defined in `package.json` after build is done.

electron/src/lib/CoreProtocol.ts
pwnsdx: It's not for developer, it's for everyone

info.json
pwnsdx: This would need to go in another PR.

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: Would be ideal but the renderer cannot access the main scope.

electron/src/lib/AutomatedSingleSignOn.ts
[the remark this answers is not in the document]
bennyn: Is this time based on laptop performance and/or network speed? Let's try to avoid timings. Webapp now supports to login via URL: https://github.com/wireapp/wire-webapp/pull/5739

electron/src/lib/AutomatedSingleSignOn.ts
bennyn: But they both could import the same JS file, can't they?

electron/src/lib/AutomatedSingleSignOn.ts
bennyn: Changes are there: https://github.com/wireapp/wire-webapp/pull/5739 (/cc @marcoconti83)

electron/src/lib/CoreProtocol.ts
bennyn: Why do we need to have 2 places to store that information?

electron/src/lib/CoreProtocol.ts
bennyn: But is there a requirement from a customer who asks for this? Wrapper can be started with not listing the protocol as first parameter and then this statement would be screwed, wouldn't it?

info.json
bennyn: I agree.

marcoconti83:
After discussing this with @bennyn, we decided to go for the '#' solution for now, where the code is passed by URL with the hash. 

This is needed for the wrapper even if we don't cater for the webapp at the moment, because the webapp _inside_ the wrapper should use the URL parameter approach. The current approach where we copy the code to the clipboard, read it, then restore the clipboard sounds way more hacky, and it has the extra disadvantage of not being able to inform the user properly if a code is invalid.

electron/src/lib/AutomatedSingleSignOn.ts
bennyn: Changes are implemented (https://github.com/wireapp/wire-webapp/pull/5752). 

@pwnsdx Link schema is now: http://localhost:8081/auth/#sso/wire-26233198-77f3-4aac-97ad-03de41168ab7 - Can you please adjust this PR here to use that?

electron/src/lib/AutomatedSingleSignOn.ts
bennyn: When using hash URLs on webapp then it can log you in automatically (if the code is correct):
- https://github.com/wireapp/wire-webapp/pull/5753

Do we still need `this.clickOn('a[data-uie-name="go-sign-in-sso"]')` when the webapp logs you in automatically?

electron/src/lib/AutomatedSingleSignOn.ts
bennyn: @pwnsdx Which element is clicked in that case? Can we make use of the new webapp functionality which logs you in automatically? https://github.com/wireapp/wire-webapp/pull/5753

electron/src/lib/CoreProtocol.ts
pwnsdx: Windows pass the wire:// URL as a second argument so if the app is not running, it will go in that specific case.

electron/src/lib/CoreProtocol.ts
pwnsdx: One is for build time, the other at runtime

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: To investigate where the file would be

electron/src/lib/CoreProtocol.ts
bennyn: Ok, makes sense. Thanks for explaining!

electron/src/lib/CoreProtocol.ts
bennyn: Can one be replaced with the other?

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: This has been removed.

electron/src/lib/AutomatedSingleSignOn.ts
pwnsdx: This has been removed.

electron/src/lib/AutomatedSingleSignOn.ts
bennyn: You are free to choose! :)

electron/src/lib/CoreProtocol.ts
bennyn: Thanks for the detailed explanation! Will mark this discussion as resolved. 👍

PR #2186, opened 2019-02-12
refactor: Unify logger settings

bennyn:


electron/src/js/getLogger.ts
```diff
@@ -0,0 +1,20 @@
+import {LogFactory, LoggerOptions} from '@wireapp/commons';
```
ffflorian: License header is missing

bennyn: Good catch! It's in now: https://github.com/wireapp/wire-desktop/commit/220497e81c0b3bb9e187117d10d0498294845e2d

PR #2193, opened 2019-02-13
runfix: Properly lock account deletion process (WEBAPP-5921)

pwnsdx:


ffflorian:
Please rename to `runfix:`

PR #2215, opened 2019-02-15
chore: Electron code coverage setup

bennyn:
Let me describe the code coverage workflow which makes it possible to get the coverage metrics for code that runs inside Electron's main process:

1. TypeScript source code gets compiled with [tsc](https://github.com/Microsoft/TypeScript)
2. Compiled code gets instrumented with [istanbul](https://github.com/gotwarlost/istanbul)
3. Instrumented code will be tested using [electron-mocha](https://github.com/jprichardson/electron-mocha) within Electron's main process
4. A script bound to mocha's global "after" hook will write the coverage info into a `coverage.json` file
5. [remap-istanbul](https://github.com/SitePen/remap-istanbul) will create a nice HTML report based on `coverage.json` and remap the compiled code to it's TypeScript sources within the report
6. Bonus: Code coverage thresholds will be checked using `istanbul check-coverage`

Although the workflow sounds straight forward, there are some caveats:

1. Within the tests we need to import the instrumented code and not the compiled code → https://github.com/jprichardson/electron-mocha/issues/19#issuecomment-193374439
2. To get proper coverage reports of the full code base, we need to require every code that is being used which is done in `BeforeTest.js` using mocha's `before` hook → https://github.com/jprichardson/electron-mocha/issues/135

Special adjustments to make it fit with our codebase:

1. Tests will be run from the `./electron` subdirectory because we need the node modules from there plus the references to `./electron/locale`
2. To prevent tests in Electron's main process from crashing when loading code that needs to be run in Electron's renderer process, I moved all renderer-related code into `./electron/src/renderer` and excluded these fragments in `BeforeTest.js`
3. I moved the configuration for `electron-mocha` into [mocha.opts](https://mochajs.org/#mochaopts), which is the default file to store configurations for mocha


electron/package.json
```diff
@@ -58,6 +61,12 @@
     "url": "https://github.com/wireapp/wire-desktop.git"
   },
   "scripts": {
+    "coverage:check": "istanbul check-coverage --branches=10 --functions=10 --lines=35 --statements=30 instrumented-code/coverage.json",
+    "coverage:clean": "rimraf coverage instrumented-code",
+    "coverage:instrument": "istanbul instrument dist --no-compact --complete-copy --output instrumented-code",
+    "coverage:report": "remap-istanbul --input instrumented-code/coverage.json --output coverage -t html",
+    "coverage:test": "electron-mocha",
```
atomrc: maybe this script can be renamed just `test`. It is not directly linked to coverage, am I wrong?

bennyn: Ok, has been renamed: https://github.com/wireapp/wire-desktop/pull/2215/commits/6f4f2e6bf583c6f320948b01603936ecfaad210b

PR #2259, opened 2019-02-26
feat: Activate logging inside of production webapp

bennyn:
When starting the wrapper with `--enable-logging` then it tells the webapp to also activate logging.


electron/src/env/LoggerUtil.ts
```diff
@@ -0,0 +1,9 @@
+function enableLogging(): boolean {
```
ffflorian:
```suggestion
/*
 * Wire
 * Copyright (C) 2019 Wire Swiss GmbH
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see http://www.gnu.org/licenses/.
 *
 */
 
function enableLogging(): boolean {
```

electron/src/main.ts
```diff
@@ -173,7 +174,11 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
   checkConfigV0FullScreen(mainWindowState);
 
   const baseURL = `${BASE_URL}${BASE_URL.includes('?') ? '&' : '?'}hl=${locale.getCurrent()}`;
-  main.loadURL(`${fileUrl(INDEX_HTML)}?env=${encodeURIComponent(baseURL)}`);
+  let url = `${fileUrl(INDEX_HTML)}?env=${encodeURIComponent(baseURL)}`;
+  if (enableLogging) {
```
ffflorian:
```suggestion
  if (enableLogging()) {
```

PR #2266, opened 2019-02-28
refactor: Spawn squirrel asynchronously

ffflorian:


electron/src/main.ts
```diff
@@ -424,7 +424,7 @@ registerCoreProtocol();
 initRaygun.initClient();
 appInit.handlePortableFlags();
 lifecycle.checkSingleInstance();
-lifecycle.checkForUpdate();
+lifecycle.checkForUpdate().catch(error => logger.error(error));
```
bennyn: Will it also work when just calling `lifecycle.checkForUpdate().catch(logger.error)`?

PR #2269, opened 2019-02-28
refactor: Use axios for link previews

ffflorian:
This closes https://github.com/wireapp/wire-desktop/pull/2271


electron/src/lib/openGraph.ts
```diff
@@ -31,75 +31,67 @@ const bufferToBase64 = (buffer: string, mimeType?: string): string => {
...
-      currentDownloadSize += buffer.length;
-      if (currentDownloadSize > IMAGE_SIZE_LIMIT) {
-        imageRequest.abort();
-        resolve();
-      }
-    });
-  });
+  const axiosConfig: AxiosRequestConfig = {
+    headers: {
+      ['User-Agent']: USER_AGENT,
```
bennyn: Do we need the brackets here?

electron/src/lib/openGraph.ts
```diff
@@ -17,91 +17,100 @@
...
 
-    getContentRequest.on('data', buffer => {
-      const chunk = buffer.toString();
+      response.data.on('data', buffer => {
+        const chunk = buffer.toString();
+        partialBody += chunk;
 
-      partialBody += chunk;
+        if (chunk.match('</head>') || partialBody.length > contentLimit) {
+          resolve(partialBody);
```
atomrc: I think you should also abort the request. Else the browser will continue downloading content

PR #2295, opened 2019-03-06
chore: Split other Linux build tasks as well

maximbaz:
No functional changes, repeating #2228 for other linux tasks. I only need it for internal (for wire-desktop-beta), but thought why not align all linux tasks at once while I'm at it.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -585,11 +585,7 @@ module.exports = function(grunt) {
 
   grunt.registerTask('linux-custom', ['linux-custom-package', 'electronbuilder:linux_prod']);
 
-  grunt.registerTask('linux-other-internal-package', ['linux-package']);
+  grunt.registerTask('linux-other-internal', ['linux-package', 'electronbuilder:linux_other']);
 
-  grunt.registerTask('linux-other-internal', ['linux-other-internal-package', 'electronbuilder:linux_other']);
-
-  grunt.registerTask('linux-other-package', ['linux-prod-package']);
-
-  grunt.registerTask('linux-other', ['linux-other-package', 'electronbuilder:linux_other']);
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -552,16 +552,17 @@ module.exports = function(grunt) {
...
     'clean:wrap',
     'update-keys',
     'gitinfo',
     'set-custom-data',
     'release-internal',
     'bundle',
-    'electronbuilder:linux_internal',
   ]);
 
+  grunt.registerTask('linux', ['linux-package', 'electronbuilder:linux_internal']);
```
ffflorian: Let's move this above the other short tasks, too.

maximbaz: You mean below, right? Done

Gruntfile.js
```diff
@@ -571,35 +570,26 @@ module.exports = function(grunt) {
...
-  ]);
+  grunt.registerTask('linux-prod', ['linux-prod-package', 'electronbuilder:linux_prod']);
+
+  grunt.registerTask('linux-custom', ['linux-custom-package', 'electronbuilder:linux_prod']);
+
+  grunt.registerTask('linux-other-internal-package', ['linux-package']);
+
+  grunt.registerTask('linux-other-internal', ['linux-other-internal-package', 'electronbuilder:linux_other']);
+
+  grunt.registerTask('linux-other-package', ['linux-prod-package']);
```
maximbaz: Note that `linux-other-internal-package` and `linux-other-package` are now simply aliases to other tasks, I didn't remove them because I don't know if you use them or not, but they are potential candidates for removal.

maximbaz: No silly me, of course the only usage is right here because `-package` didn't exist before 😄  Updated.

PR #2316, opened 2019-03-11
refactor: Replace winston logger with LogFactory

bennyn:


electron/src/main.ts
```diff
@@ -316,6 +316,12 @@ const initElectronLogFile = (): void => {
   fs.ensureFileSync(LOG_FILE);
 };
 
+const getWebViewId = (contents: Electron.WebContents): string => {
+  const url = contents.getURL();
+  const search = '&id=';
```
pwnsdx: Please use URL module from Node.JS and the searchParams

electron/src/main.ts
```diff
@@ -387,6 +393,17 @@ class ElectronWrapperInit {
           // Open webview links outside of the app
           contents.on('new-window', openLinkInNewWindow);
           contents.on('will-navigate', willNavigateInWebview);
+          contents.on('console-message', async (event, level, message) => {
+            const webViewId = getWebViewId(contents);
+            if (webViewId) {
+              const logFilePath = path.join(app.getPath('userData'), 'logs', webViewId, config.LOG_FILE_NAME);
```
pwnsdx: We need to make sure the UUID is a valid one so we need to check it

PR #2347, opened 2019-03-15
refactor: Restructure "js" directory

bennyn:
This PR gets rid off the "js" directory. It also disables hard code coverage thresholds because we cannot combine coverage reports from main & renderer process yet (see https://github.com/istanbuljs/nyc/issues/1011).


electron/src/util/OriginValidator.ts
```diff
@@ -0,0 +1,9 @@
+import * as url from 'url';
```
ffflorian: Missing copyright header

electron/src/window/AboutWindow.ts
```diff
@@ -142,4 +147,4 @@ const showWindow = () => {
   aboutWindow.show();
 };
 
-export {showWindow};
+export default {showWindow};
```
ffflorian: I thought we agreed to not use `export default` anymore :wink:

electron/src/window/AboutWindow.ts
```diff
@@ -18,13 +18,13 @@
  */
 
 import {BrowserWindow, IpcMessageEvent, app, ipcMain, session, shell} from 'electron';
-import fileUrl = require('file-url');
 import * as path from 'path';
 
-import {i18nLanguageIdentifier} from '../interfaces/';
+import fileUrl = require('file-url');
```
ffflorian: This should work, too:

```suggestion
import fileUrl from 'file-url';
```

bennyn:
Thanks! Fixed.

PR #2354, opened 2019-03-18
refactor: Remove grunt

ffflorian:
Breaking changes:
- all builds are now in `wrap/dist` (instead of `wrap/prod`, `wrap/internal`, ...)
- `info.json` is now `wire.json`
- `BUILD_ENV` is now `APP_ENV`


electron/src/settings/config.ts
```diff
@@ -17,17 +17,22 @@
...
+  raygunApiKey: string;
+  version: string;
+} = require('../../wire.json');
 
 const BACKEND_ORIGINS = ['https://staging-nginz-https.zinfra.io', 'https://prod-nginz-https.wire.com'];
 
 const LOG_FILE_NAME = 'console.log';
 
-const MAXIMUM_ACCOUNTS = pkg.maximumAccounts || 3;
+const MAXIMUM_ACCOUNTS = Number(config.maximumAccounts);
```
bennyn: Let's use `parseInt`

electron/wire.json
```diff
@@ -0,0 +1,19 @@
...
+  "adminUrl": "https://teams.wire.com",
+  "appBase": "https://app.wire.com",
+  "buildNumber": "0",
+  "copyright": "© Wire Swiss GmbH",
+  "customProtocolName": "wire",
+  "description": "The most secure collaboration platform.",
+  "environment": "internal",
+  "legalUrl": "https://wire.com/legal/",
+  "licensesUrl": "https://wire.com/legal/licenses/",
+  "maximumAccounts": "3",
```
atomrc: Let's put this as a number instead of a string

ffflorian: Since it will be injected as a string when loading a configuration via `process.env`, I figured we should save it as a string here, too.

PR #2373, opened 2019-03-22
fix: Revert Squirrel related changes

ffflorian:
This PR reverts major changes in 5b2e1af, 1ca3654 and d43b11d and keeps only introduced fixes.


electron/src/update/Squirrel.ts
```diff
@@ -17,11 +17,15 @@
  *
  */
 
-import {spawn} from 'child_process';
+// https://github.com/atom/atom/blob/master/src/main-process/squirrel-update.js
```
bennyn: Let's link to a specific commit rather than to the "master" branch (which might change over time).

PR #2375, opened 2019-03-22
feat: Show conversation via custom protocol link

bennyn:


electron/src/lib/CoreProtocol.ts
```diff
@@ -44,12 +50,19 @@ const dispatcher = async (url?: string) => {
...
   switch (route.host) {
-    case CORE_PROTOCOL_SSO: {
+    case ProtocolCommand.START_SSO_FLOW: {
       logger.log('Automatic SSO detected');
       await AutomatedSingleSignOn.handleProtocolRequest(route);
       break;
     }
-
+    case ProtocolCommand.SHOW_CONVERSATION: {
+      const primaryWindow = WindowManager.getPrimaryWindow();
```
Yserz: Question: What happens if there are 2 accounts in the wrapper and both have the possibility to open the conversation with the given ID? Are we only considering the currently active account (PrimaryWindow?)?

bennyn: Yes, we will only take care off the main active account.

PR #2392, opened 2019-03-26
refactor: Reuse logic to send messages to renderer processes

bennyn:
We are communicating from Electron's main process (which gets information from custom protocol links) to the renderer process by using `main.webContents.send`. This works fine, when the wrapper is already up and running. However, it does not work when the wrapper is closed and needs to be started first because our WebViews mights not be in a state where they can receive asynchronous messages. That's why I reuse the logic from @pwnsdx to only send a signal once we get a ['did-finish-load'](https://github.com/electron/electron/blob/master/docs/api/web-contents.md#event-did-finish-load) event.


electron/src/lib/AutomatedSingleSignOn.ts
```diff
@@ -17,40 +17,13 @@
...
-      main.webContents.once('did-finish-load', () => sendCodeToRenderer());
-    } else {
-      if (!main.isVisible()) {
-        main.show();
-        main.focus();
-      }
-      sendCodeToRenderer();
-    }
-  };
+const dialog = require('electron').dialog || require('electron').remote.dialog;
```
ffflorian:
```suggestion
import {dialog: mainDialog, remote} from 'electron'
const dialog = mainDialog || remote.dialog;
```

electron/src/lib/CoreProtocol.ts
```diff
@@ -53,13 +52,20 @@ const dispatcher = async (url?: string) => {
     case ProtocolCommand.SHOW_CONVERSATION: {
       const conversationIds = route.pathname.match(ValidationUtil.PATTERN.UUID_V4);
       if (conversationIds) {
-        WindowManager.sendActionToPrimaryWindow(EVENT_TYPE.CONVERSATION.SHOW, conversationIds[0]);
+        const conversationId = conversationIds[0];
+        logger.log(`Show conversation "${conversationId}"...`);
```
```suggestion
        logger.log(`Showing conversation "${conversationId}"...`);
```

PR #2394, opened 2019-03-26
fix: Make deep links work on startup (WEBAPP-6026)

bennyn:


electron/src/lib/CoreProtocol.ts
```diff
@@ -36,61 +36,74 @@ const CORE_PROTOCOL_MAX_LENGTH = 1024;
...
 }
 
-const dispatcher = async (url?: string) => {
-  if (typeof url === 'undefined' || !url.startsWith(`${CORE_PROTOCOL}://`) || url.length > CORE_PROTOCOL_MAX_LENGTH) {
-    return;
-  }
+export class CustomProtocolHandler {
+  public hashLocation: string = '';
+
+  async dispatcher(url?: string) {
```
atomrc: I would rename this function to `dispatch` instead of `dispatcher` for semantic purposes :)

electron/src/main.ts
```diff
@@ -184,6 +185,10 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
     webappURL += `&enableLogging=@wireapp/*`;
   }
 
+  if (customProtocolHandler.hashLocation) {
+    webappURL += `#${customProtocolHandler.hashLocation}`;
```
ffflorian: What do you think about using Node's [`url`](https://nodejs.org/api/url.html) module to build the URL? Constructing the URL is getting a bit fiddly and I am also wondering if it works with `encodeURIComponent()` in line 196.

atomrc: It's probably a good idea, indeed, to use node's url builder :+1: 
There should be no problem with the `encodeUIRComponent`, we still need that as it is passed as a url parameter

bennyn: @ffflorian I will check if we can get rid off the `file-url` package which is being used here.

PR #2417, opened 2019-03-29
refactor: Convert JavaScript tests to TypeScript

bennyn:
This PR converts our JavaScript tests into TypeScript. 🎉 

It brings the big benefit that we can import now the real source code in our tests rather than importing the instrumented and compiled code.

There are some noteworthy things though:

- Code coverage will be removed and added back later; we will get a much cleaner solution where we don't need to take care about the instrumentation of our code
- `import * as minimist from 'minimist'` needed to be turned into a `require` statements because the `@babel/preset-typescript` doesn't liked the `import` syntax for it (it also does not like `import minimist = require('minimist')`)
- The `\"**/*.ts\"` file pattern has been removed from the `tslint` script because we use already `--project` which defines the files that should be linted by relying on `tsconfig.json` (see https://github.com/palantir/tslint/issues/4086#issuecomment-409745298)


electron/babel-register.js
```diff
@@ -0,0 +1,16 @@
+require('@babel/register')({
```
ffflorian:
```suggestion
const babelRegister = require('@babel/register');
babelRegister({
```

electron/src/menu/TrayHandler.test.main.ts
```diff
@@ -0,0 +1,115 @@
+import sinon from 'sinon';
```
ffflorian: Missing license header

electron/src/menu/TrayHandler.test.main.ts
```diff
@@ -0,0 +1,115 @@
...
+import * as assert from "assert";
```
ffflorian:
```suggestion
import * as assert from 'assert';
```

electron/src/menu/TrayHandler.test.main.ts
```diff
@@ -0,0 +1,115 @@
...
+import * as path from 'path';
+import {TrayHandler} from "./TrayHandler";
```
ffflorian:
```suggestion
import {TrayHandler} from './TrayHandler';
```

package.json
```diff
@@ -54,17 +55,22 @@
...
+        "tslint --config tslint.json --project tsconfig.json --fix",
+        "git add"
+      ],
+      "*.{json,md,css}": [
+        "prettier --write",
+        "git add"
+      ]
+    },
+    "ignore": [
+      "electron/**/*.test*.ts"
```
ffflorian: Why ignore them?

electron/src/sso/SingleSignOn.test.main.ts
```diff
@@ -0,0 +1,10 @@
+import * as assert from "assert";
```
ffflorian: Missing license header
and
```suggestion
import * as assert from 'assert';
```

package.json
bennyn: Because the linter uses our main `tsconfig.json` when linting and the test files are excluded from our main `tsconfig.json` and the linter will then complain about that but the test files actually don't need to be included in our main `tsconfig.json` because they are piped through the Babel TypeScript compiler which has it's own compatibility configuration.

ffflorian: Then let's create a `tsconfig.jasmine.json` so that `tsconfig` can fix them on commit.

electron/src/runtime/EnvironmentUtil.ts
```diff
@@ -42,7 +42,7 @@ enum BackendType {
   RC = 'RC',
 }
 
-enum BackendTypeLabel {
+export enum BackendTypeLabel {
```
atomrc: What is the difference with the previous version of the export?

electron/test/main/SingleSignOnTest.js
```diff
@@ -1,10 +0,0 @@
-const assert = require('assert');
```
atomrc: This one seems to not have been translated to TS

electron/src/runtime/EnvironmentUtil.ts
bennyn: The previous one didn't compile with `@babel/preset-typescript`.

electron/test/main/SingleSignOnTest.js
bennyn: Mmmmm. I deleted that file. 🐺

package.json
bennyn: Seems like we can go without it.

PR #2419, opened 2019-03-29
test: Add tests for deep links

bennyn:


electron/src/lib/CoreProtocol.test.main.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('forwards conversation deep links to the webapp', () => {
+    sinon.replace(protocolHandler, 'forwardHashLocation', sinon.fake());
+    protocolHandler.dispatcher('wire://conversation/8cdb44a0-418b-4188-9a53-7c477a7848dd');
+    assert.strictEqual((protocolHandler['forwardHashLocation'] as sinon.SinonSpy).calledOnce, true);
+  });
+
+  it('forwards user profile deep links to the webapp', () => {
+    sinon.replace(protocolHandler, 'forwardHashLocation', sinon.fake());
```
atomrc: Mocking private function means you are not testing most of the logic.
What you want to mock would rather be the `WindowManager.sendActionToPrimaryWindow` method and check that it's really called.

bennyn: Yepp. I can mock `sendActionToPrimaryWindow` and do an expectation on the parameters that have been used to call it. Good point. Will mark my PR as draft.

electron/src/lib/CoreProtocol.test.main.ts
```diff
@@ -0,0 +1,33 @@
+import * as assert from 'assert';
```
ffflorian: Missing license header

electron/src/lib/CoreProtocol.ts
```diff
@@ -49,26 +50,30 @@ export class CustomProtocolHandler {
...
+      logger.log('Starting SSO flow...');
+      const code = route.pathname.trim().substr(1);
+      this.windowManager.sendActionAndFocusWindow(EVENT_TYPE.ACCOUNT.SSO_LOGIN, code).catch(error => {
+        logger.error(`Cannot start SSO flow: ${error.message}`, error);
+      });
     }
   }
 
-  public registerCoreProtocol() {
+  registerCoreProtocol() {
```
ffflorian: Everywhere else we are using the `public` keyword. Why remove it now? I think it adds value to the code since it makes it more clear that this function or variable is actually public.

electron/src/lib/CoreProtocol.test.main.ts
```diff
@@ -0,0 +1,33 @@
...
+import sinon from 'sinon';
+import {CustomProtocolHandler} from './CoreProtocol';
+import {EVENT_TYPE} from "./eventType";
```
ffflorian:
```suggestion
import {EVENT_TYPE} from './eventType';
```

electron/src/lib/CoreProtocol.ts
bennyn: I think we can get rid off the `public` keyword because it's the default.

ffflorian: Yes, we could, but as I said, I think it adds more value to the code.
Explicit > implicit IMO

electron/src/lib/CoreProtocol.ts
```diff
@@ -49,26 +50,30 @@ export class CustomProtocolHandler {
...
+      await this.handleSSOLogin(route);
+    } else {
+      this.forwardHashLocation(route);
+    }
+  }
+
+  private forwardHashLocation(route: URL): void {
+    const location = route.href.substr(CORE_PROTOCOL_PREFIX.length);
+    this.hashLocation = `/${location}`;
+    this.windowManager.sendActionToPrimaryWindow(EVENT_TYPE.WEBAPP.CHANGE_LOCATION_HASH, this.hashLocation);
```
```suggestion
    await this.windowManager.sendActionToPrimaryWindow(EVENT_TYPE.WEBAPP.CHANGE_LOCATION_HASH, this.hashLocation);
```

electron/src/lib/CoreProtocol.ts
```diff
@@ -49,26 +50,30 @@ export class CustomProtocolHandler {
...
+  }
+
+  private handleSSOLogin(route: URL): Promise<void> {
+    if (typeof route.pathname === 'string') {
+      logger.log('Starting SSO flow...');
+      const code = route.pathname.trim().substr(1);
+      this.windowManager.sendActionAndFocusWindow(EVENT_TYPE.ACCOUNT.SSO_LOGIN, code).catch(error => {
```
```suggestion
      try {
 
       await this.windowManager.sendActionAndFocusWindow(EVENT_TYPE.ACCOUNT.SSO_LOGIN, code);
      } catch(error) {
```

electron/src/lib/CoreProtocol.test.main.ts
bennyn: Reminds me to update the linting script...

electron/src/lib/CoreProtocol.test.main.ts
bennyn: My favorite error 🗡. I should make more use of my IDE: https://www.jetbrains.com/help/webstorm/copyright.html

electron/src/lib/CoreProtocol.ts
bennyn: I agree that the function definitely misses a return type because this one cannot be easily guessed but I would like to avoid defining things as `public` (coding by convention).

electron/src/lib/CoreProtocol.test.main.ts
ffflorian: Maybe merge from staging first 😉

electron/src/lib/CoreProtocol.test.main.ts
```diff
@@ -0,0 +1,33 @@
...
+
+let protocolHandler: CustomProtocolHandler = undefined;
+
+describe('dispatcher', () => {
```
atomrc: One last thing, let's rename this function `dispatch` instead of `dispatcher`. WDYT?

bennyn: I would give it an even more precise name, like `dispatchDeepLink`.

atomrc: Sure :+1:

electron/src/lib/CoreProtocol.test.main.ts
bennyn: I see what you did there (https://github.com/wireapp/wire-desktop/pull/2426) but we have "mocha" here, not "jest". ;)

ffflorian: Fixed in https://github.com/wireapp/wire-desktop/pull/2433. I was more talking about the automatic linting on commit.

electron/src/lib/CoreProtocol.ts
```diff
@@ -40,52 +40,59 @@ enum ProtocolCommand {
...
+    }
+  }
+
+  private forwardHashLocation(route: URL): void {
+    const location = route.href.substr(CORE_PROTOCOL_PREFIX.length);
+    this.hashLocation = `/${location}`;
+    this.windowManager.sendActionToPrimaryWindow(EVENT_TYPE.WEBAPP.CHANGE_LOCATION_HASH, this.hashLocation);
+  }
+
+  private async handleSSOLogin(route: URL): void {
```
```suggestion
  private async handleSSOLogin(route: URL): Promise<void> {
```

electron/src/lib/CoreProtocol.ts
```diff
@@ -40,52 +40,59 @@ enum ProtocolCommand {
...
+      try {
+        await this.windowManager.sendActionAndFocusWindow(EVENT_TYPE.ACCOUNT.SSO_LOGIN, code);
+      } catch (error) {
+        logger.error(`Cannot start SSO flow: ${error.message}`, error);
       }
     }
   }
 
-  public registerCoreProtocol() {
+  public async registerCoreProtocol(): void {
```
```suggestion
  public async registerCoreProtocol(): Promise<void> {
```

bennyn: Why do you want to declare this one as `async`?

bennyn: We should remove `async` for `registerCoreProtocol` because it's not `async`. Only the callbacks from the listeners that it creates are async.

bennyn:
> Why use finally? It will try to assert even when line 53 fails.

When line 53 fails, then the assert will also fail. That way I don't need to catch the error and throw an assertion error myself.

PR #2475, opened 2019-04-09
chore: Don't archive rpm builds, archive sha256sum

ffflorian:


bin/linux-checksums.sh
```diff
@@ -61,22 +61,12 @@ if ! ls ./*.deb > /dev/null 2>&1; then
...
   _error_exit "No AppImage files found. Add some in ${PWD}."
 fi
 
 _log "Creating checksums..."
-sha256sum *.deb *.rpm *.AppImage > sha256sum.txt
-
-_log "Creating source code archive for signing..."
-(
-  cd ../../
-  git archive -o "wrap/dist/${BUILD_VERSION}.tar.gz" --format tar.gz --prefix "wire-desktop-linux-${BUILD_VERSION}/" master
```
mythsunwind: Why no source code archive anymore? Isn't this used for github releases?

ffflorian: It was never working properly and we should either find a reliable solution or do it manually. I always did it manually for the last releases.

ffflorian: Actually, I can create a bash script for that :nerd_face:

PR #2497, opened 2019-04-12
fix: Correctly log variables

ffflorian:


electron/src/sso/SingleSignOn.ts
```diff
@@ -210,15 +210,10 @@ class SingleSignOn {
...
-    this.session.webRequest.onBeforeSendHeaders(
-      {
-        urls: ['*'],
-      },
-      (details: any, callback: Function) => {
-        details.requestHeaders['User-Agent'] = config.userAgent;
-        callback({cancel: false, requestHeaders: details.requestHeaders});
-      }
-    );
+    this.session.webRequest.onBeforeSendHeaders({urls: ['*']}, (details: any, callback: Function) => {
```
ffflorian: Formatting only here

PR #2514, opened 2019-04-16
test: Setup test environment for renderer process

bennyn:


electron/src/renderer/menu/preload-about.ts
```diff
@@ -30,14 +30,13 @@ ipcRenderer.once(EVENT_TYPE.ABOUT.LOCALE_RENDER, (event: IpcMessageEvent, labels
...
 
 interface Details {
-  copyright: string;
-  electronVersion: string;
-  environment: string;
+  copyright?: string;
+  electronVersion?: string;
   productName: string;
-  webappVersion: string;
+  webappVersion?: string;
```
ffflorian: Why did you change this interface? It now doesn't reflect the data sent in [`AboutWindow.ts`](https://github.com/wireapp/wire-desktop/blob/staging/electron/src/window/AboutWindow.ts) anymore.

bennyn: I changed it because `Details` is only used in `preload-about.ts` and within this file it's optional:

**Usage**

```ts
    if (details.webappVersion) {
      webappVersionElement.innerHTML = details.webappVersion;
    }
```

ffflorian: But IMO the interface should reflect the original data, not the way the data is used in this file.

bennyn: @ffflorian Do you think we can remove the existing `if (details.webappVersion)` check?

ffflorian: Yes, we can remove that. :slightly_smiling_face:

bennyn: Ok, then I will remove the `else` case too and mark `webappVersion` as mandatory.

bennyn: Changed: https://github.com/wireapp/wire-desktop/pull/2514/commits/8219e3aa8cc277529c58627e8e9d63fdd93751cc

ffflorian: But what about the rest of the interface?

bennyn: Let's take `Details.electronVersion` as an example. I marked it as optional because there is code that expects it not to be present:

**Example**

```ts
versionElement.innerHTML = details.electronVersion || 'Development';
```

This clearly shows that `electronVersion` is expected to be optional so I reflected this in the interface. 

If we want to make `electronVersion` mandatory, then we need to remove the boolean OR because it then makes no sense. My initial plan with this PR was to avoid changes in the behaviour of the functional code.

PR #2519, opened 2019-04-17
runfix: Don't copy config files from node_modules

ffflorian:
Otherwise builds with completely different config repos set via [`WIRE_CONFIGURATION_REPOSITORY`](https://github.com/wireapp/wire-web-packages/blob/master/packages/copy-config/README.md#environment-variables) will never copy the correct files but always the default files.

This partly reverts https://github.com/wireapp/wire-desktop/pull/2512.


Yserz:
😢

Issue #2097, opened 2019-01-17
Link Preview doesn't honors page encoding


Snawoot:
Wire version: 3.5.2881
Operating system: Debian 9
Which antivirus software do you have installed: None

What steps will reproduce the problem?

1.  Send a link pointing to non-unicode page (like this one: https://www.opennet.ru/opennews/art.shtml?num=49969)

What is the expected result?

Normal preview of page caption

What is the actual result?

Preview without respect of non-unicode page encoding. Text looks like bunch of question marks.

Please provide any additional information below. Attach a screenshot if possible.

Preview looks like this:
![Screenshot](https://gist.githubusercontent.com/Snawoot/8e21d3404f1708ba05232dc984bc943b/raw/7b0ddb05c3d9423c1e249c660faa4e90da6d3676/2019-01-17-115605_464x104_scrot.png)

pitachx:
+1

thisnickwasfree:
Can confirm this. Any link from opennet.ru, pikabu.ru has wrong encoding in wire's preview.

PR #2853, opened 2019-07-18
fix: Decode non-UTF-8 websites

ffflorian:
This PR will
- detect the encoding of a website and decode it with [iconv-lite](https://www.npmjs.com/package/iconv-lite) to properly display link previews

---

Example (before and after):

screenshot

This fixes https://github.com/wireapp/wire-desktop/issues/2097.


electron/src/lib/openGraph.ts
```diff
@@ -80,15 +89,24 @@ const axiosWithContentLimit = (config: AxiosRequestConfig, contentLimit: number)
...
-        const isHtmlContentType = contentType.match(/.*text\/html/);
+        let contentType;
+
+        try {
+          contentType = parseContentType(response.headers['content-type'] || '');
+        } catch (error) {
+          return reject(`Could not parse content type: "${error.message}"`);
+        }
+
+        const isHtmlContentType = contentType.type.match(/.*text\/html/);
```
AndyLnd:
```suggestion
        const isHtmlContentType = /text\/html/.test(contentType.type);
```

PR #2918, opened 2019-08-06
fix: Create link previews for websites with invalid head (WEBAPP-6278)

ffflorian:
A website consisting only of 
```html
<head>
  <meta property="og:image" content="http://example.com/image.jpg">
```
is technically not valid HTML5 **but** it has a preview image. Browsers will also display it correctly.

### DIFF
--- a/electron/src/lib/openGraph.ts
+++ b/electron/src/lib/openGraph.ts
@@ -148,7 +148,7 @@ const fetchOpenGraphData = async (url: string): Promise<OpenGraphResult> => {
   };
 
   const body = await axiosWithContentLimit(axiosConfig, CONTENT_SIZE_LIMIT);
-  const matches = body.match(/.*property="og:[^"]+".*/gim) || [''];
+  const matches = body.match(/.*property=(["'])og:.+?\1.*/gim) || [''];
 
   if (!matches) {
     throw new Error('No open graph tags found in website.');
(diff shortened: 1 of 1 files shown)


electron/src/lib/openGraph.ts
```diff
@@ -146,13 +146,15 @@ const fetchOpenGraphData = async (url: string): Promise<OpenGraphResult> => {
   };
 
   const body = await axiosWithContentLimit(axiosConfig, CONTENT_SIZE_LIMIT);
-  const [head] = body.match(/<head>[\s\S]*?<\/head>/) || [''];
+  const matches = body.match(/.*property="og:[^"]+".*/gim) || [''];
```
ffflorian: https://regex101.com/r/kwrYZD/1

electron/src/lib/openGraph.ts
```diff
@@ -42,6 +42,8 @@ const bufferToBase64 = (buffer: Buffer, mimeType: string): string => {
 
 const fetchImageAsBase64 = async (url: string): Promise<string | undefined> => {
   const IMAGE_SIZE_LIMIT = 5e6; // 5MB
+  const parsedUrl = parseUrl(encodeURI(url));
+  const normalizedUrl = parsedUrl.protocol ? parsedUrl : parseUrl(`http://${url}`);
```
lipis: https? or this for the ones that don't have the protocol?

electron/src/lib/openGraph.ts
AndyLnd: what about single quotes?
```suggestion
  const matches = body.match(/.*property=(["'])og:.+?\1.*/gim) || [''];
```

electron/src/lib/openGraph.ts
ffflorian: This is for the ones that don't have a protocol.

PR #2937, opened 2019-08-14
refactor: Remove unnecessary filters, better types

ffflorian:


electron/src/sso/SingleSignOn.ts
```diff
@@ -213,16 +213,19 @@ export class SingleSignOn {
     this.session.setPermissionRequestHandler((webContents, permission, callback) => callback(false));
 
     // User-agent normalization
-    this.session.webRequest.onBeforeSendHeaders({urls: ['*']}, (details: any, callback: Function) => {
-      details.requestHeaders['User-Agent'] = config.userAgent;
+    this.session.webRequest.onBeforeSendHeaders({urls: ['*']}, (details, callback) => {
+      details.requestHeaders = {
```
atomrc: Do you really need to put it back into the `details` object? (I mean mutate this object).
Since it's given to the callback, it might not be needed. WDYT?

ffflorian: The problem is, when I write it as 
```ts
details.requestHeaders['User-Agent'] = config.userAgent;
```
TypeScript won't allow it since the `RequestHeaders` in Electron is empty:

```ts
interface RequestHeaders {
}
```
and it doesn't expect `'User-Agent'` in there:

```
Element implicitly has an 'any' type because expression of type '"User-Agent"' can't be used to index type 'RequestHeaders'.
Property 'User-Agent' does not exist on type 'RequestHeaders'. ts(7053)
```

Anyway, I found out that Electron is using `any` there, too:
https://github.com/electron/electron/blob/cfd230d7f1905973113ded0b0b9965dbc4ad46b4/spec/ts-smoke/electron/main.ts#L1173

:roll_eyes:

atomrc: Oh no, what I meant was:

```
const headers = { /*...*/}

callback({ ..., requestHeaders: headers });
```

instead of mutation the `details` object (but maybe Electron relies on this mutation, that would be ugly but, well, ...

ffflorian: I'm going to use `any` for now.

electron/src/sso/SingleSignOn.ts
```diff
@@ -213,16 +213,16 @@ export class SingleSignOn {
     this.session.setPermissionRequestHandler((webContents, permission, callback) => callback(false));
 
     // User-agent normalization
-    this.session.webRequest.onBeforeSendHeaders({urls: ['*']}, (details: any, callback: Function) => {
-      details.requestHeaders['User-Agent'] = config.userAgent;
+    this.session.webRequest.onBeforeSendHeaders((details, callback) => {
```
ffflorian: See https://github.com/electron/electron/blob/v5.0.9/docs/api/web-request.md:

> If the `filter` is omitted then all requests will be matched.

electron/src/window/AboutWindow.ts
```diff
@@ -76,7 +76,7 @@ const showWindow = async () => {
     // Prevent any kind of navigation
     // will-navigate is broken with sandboxed env, intercepting requests instead
     // see https://github.com/electron/electron/issues/8841
-    aboutWindow.webContents.session.webRequest.onBeforeRequest({urls: ['*']}, async ({url}, callback) => {
+    aboutWindow.webContents.session.webRequest.onBeforeRequest(async ({url}, callback) => {
```
ffflorian: See https://github.com/electron/electron/blob/v5.0.9/docs/api/web-request.md:

> If the `filter` is omitted then all requests will be matched.

PR #3001, opened 2019-09-03
fix: Speed up link preview generation (WEBAPP-6415)

ffflorian:
Finding a regex match in a large website is very costly and can take a long time. Since open-graph [uses cheerio](https://github.com/samholmes/node-open-graph/blob/master/index.js#L59) to parse the content, it's much faster. As a result we will just send the whole body to open-graph instead of trying to parse it ourselves.

### DIFF
--- a/electron/src/lib/openGraph.ts
+++ b/electron/src/lib/openGraph.ts
@@ -167,16 +167,7 @@ const fetchOpenGraphData = async (url: string): Promise<OpenGraphResult> => {
   };
 
   const body = await axiosWithContentLimit(axiosConfig, CONTENT_SIZE_LIMIT);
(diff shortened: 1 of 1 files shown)


electron/src/lib/openGraph.ts
```diff
@@ -200,7 +191,7 @@ export const getOpenGraphData = async (url: string, callback: GetDataCallback):
       const [imageUrl] = arrayify(meta.image.url);
 
       const uri = await fetchImageAsBase64(imageUrl);
-      meta = await updateMetaDataWithImage(meta, uri);
+      meta = updateMetaDataWithImage(meta, uri);
```
ffflorian: This was unnecessary

PR #3082, opened 2019-10-02
fix: Bubble up errors from openGraph requests

ffflorian:


electron/src/lib/openGraph.ts
```diff
@@ -61,7 +61,6 @@ const fetchImageAsBase64 = async (url: string): Promise<string | undefined> => {
   try {
     response = await axiosWithCookie<Buffer>(axiosConfig);
   } catch (error) {
-    logger.error(error);
```
ffflorian: Removed this to log this only once (will be logged by the webapp later) and to avoid errors in the console which should be warnings.

electron/src/lib/openGraph.ts
```diff
@@ -150,7 +149,14 @@ export const axiosWithContentLimit = (config: AxiosRequestConfig, contentLimit:
         response.data.on('error', reject);
         response.data.on('end', () => resolve(partialBody));
       })
-      .catch(error => (axios.isCancel(error) ? Promise.resolve('') : Promise.reject(error)));
```
ffflorian: This was not using the outer `Promise` and thus created an uncaught promise rejection.

PR #3131, opened 2019-10-21
feat: Add authenticated proxy support (WEBAPP-6412) (backport)

ffflorian:


electron/src/main.ts
```diff
@@ -283,6 +304,48 @@ const handleAppEvents = () => {
...
+        logger.info('Sending provided credentials to authenticated proxy ...');
+        return callback(authenticatedProxyInfo.username, authenticatedProxyInfo.password);
+      }
+
+      const systemProxy = await getProxySettings();
+      const systemProxySettings = systemProxy && (systemProxy.http || systemProxy.https);
+
+      if (systemProxySettings) {
+        authenticatedProxyInfo = new URL(
+          `http://${systemProxySettings.credentials.username}:${systemProxySettings.credentials.password}@${authInfo.host}:${authInfo.port}`,
```
pwnsdx: We should respect the system-wide settings regarding the protocol. http is used here but it could be a https proxy. Also what about socks5://... etc?

electron/src/main.ts
```diff
@@ -283,6 +304,48 @@ const handleAppEvents = () => {
...
+        );
+        return callback(systemProxySettings.credentials.username, systemProxySettings.credentials.password);
+      }
+
+      ipcMain.once(
+        EVENT_TYPE.PROXY_PROMPT.SUBMITTED,
+        (event: IpcMessageEvent, data: {password: string; username: string}) => {
+          logger.log('Proxy prompt was submitted');
+          authenticatedProxyInfo = new URL(
+            `http://${data.username}:${data.password}@${authInfo.host}:${authInfo.port}`,
```
pwnsdx: Maybe we should ask and receive the type of proxy here as well as we seems to force the use of HTTP

electron/src/window/ProxyPromptWindow.ts
```diff
@@ -0,0 +1,103 @@
...
+      alwaysOnTop: true,
+      backgroundColor: '#ececec',
+      fullscreen: false,
+      height: windowSize.HEIGHT,
+      maximizable: false,
+      minimizable: false,
+      resizable: false,
+      show: false,
+      title: config.name,
+      webPreferences: {
```
pwnsdx: It seems we lack the `sandbox` option here. Is it intended?

electron/src/main.ts
```diff
@@ -203,6 +220,10 @@ const showMainWindow = (mainWindowState: WindowStateKeeper.State) => {
     main.webContents.openDevTools({mode: 'detach'});
   }
 
+  if (session.defaultSession) {
+    session.defaultSession.allowNTLMCredentialsForDomains('*');
```
pwnsdx: The origin of the proxy should be enough, otherwise we open this to all domains. If it's not available, maybe then we should enable it only if a proxy as been set.

electron/src/renderer/menu/preload-proxy-prompt.ts
```diff
@@ -0,0 +1,78 @@
...
+    okButton.addEventListener('click', () => sendData());
+
+    cancelButton.addEventListener('click', () => {
+      ipcRenderer.send(EVENT_TYPE.PROXY_PROMPT.CANCELED);
+      window.close();
+    });
+
+    window.addEventListener('keydown', event => {
+      if (event.key === 'Enter') {
+        sendData();
```
pwnsdx: Maybe we should detect when the form is submitted instead? Otherwise tapping on enter anywhere on the page (even if the form is not focused) will validate it

electron/src/main.ts
pwnsdx: If `authInfo.port` is not available we should rely on the protocol defaults:

```suggestion
            `http://${data.username}:${data.password}@${authInfo.host}${authInfo.port ? `:${authInfo.port}` : ''}`,
```

electron/src/window/ProxyPromptWindow.ts
ffflorian: Yes, with the `sandbox` option the translated labels are not loaded correctly.

electron/src/main.ts
ffflorian: I figured out that `authInfo.host` already includes the port.

electron/src/main.ts
```diff
@@ -283,6 +300,56 @@ const handleAppEvents = () => {
...
+
+      const systemProxy = await getProxySettings();
+      const systemProxySettings = systemProxy && (systemProxy.http || systemProxy.https);
+
+      if (systemProxySettings) {
+        const {
+          credentials: {username, password},
+        } = systemProxySettings;
+        authenticatedProxyInfo = new URL(
+          `http${systemProxy.https ? 's' : ''}://${username}:${password}@${authInfo.host}`,
```
pwnsdx: Why don't we just use `systemProxy.protocol` instead?

```suggestion
          `${systemProxy.protocol}//${username}:${password}@${authInfo.host}`,
```

electron/src/main.ts
```diff
@@ -283,6 +300,56 @@ const handleAppEvents = () => {
...
+        return callback(username, password);
+      }
+
+      ipcMain.once(
+        EVENT_TYPE.PROXY_PROMPT.SUBMITTED,
+        (event: IpcMessageEvent, promptData: {password: string; username: string}) => {
+          const {username, password} = promptData;
+
+          logger.log('Proxy prompt was submitted');
+          authenticatedProxyInfo = new URL(`http://${username}:${password}@${authInfo.host}`);
```
pwnsdx: So we are not supporting https proxies for now?

electron/src/main.ts
```diff
@@ -413,9 +480,29 @@ class ElectronWrapperInit {
       }
     };
 
-    app.on('web-contents-created', (webviewEvent: Electron.Event, contents: Electron.WebContents) => {
+    app.on('web-contents-created', async (webviewEvent: Electron.Event, contents: Electron.WebContents) => {
       WebViewFocus.bindTracker(webviewEvent, contents);
 
+      if (authenticatedProxyInfo && authenticatedProxyInfo.origin && contents.session) {
+        const proxyURL = authenticatedProxyInfo.origin;
```
pwnsdx:
```suggestion
        const proxyURL = `${authenticatedProxyInfo.protocol}//${authenticatedProxyInfo.origin}`;
```

electron/src/main.ts
ffflorian: We are now :)

PR #3144, opened 2019-10-22
fix: Register SSO protocol only once (WEBAPP-6372) (backport)

ffflorian:
Backported https://github.com/wireapp/wire-desktop/pull/2956 to Electron 4.

### DIFF
--- a/electron/src/sso/SingleSignOn.ts
+++ b/electron/src/sso/SingleSignOn.ts
@@ -148,17 +148,15 @@ export class SingleSignOn {
         }
       };
 
-      const isHandled = await new Promise<boolean>(resolve =>
-        session.protocol.isProtocolHandled(SingleSignOn.SSO_PROTOCOL, result => resolve(result as any)),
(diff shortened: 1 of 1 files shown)


electron/src/sso/SingleSignOn.ts
```diff
@@ -148,11 +148,17 @@ export class SingleSignOn {
         }
       };
 
-      session.protocol.registerStringProtocol(SingleSignOn.SSO_PROTOCOL, handleRequest, error => {
-        if (error) {
-          throw new Error(`Failed to register protocol. Error: ${error}`);
-        }
-      });
+      const isHandled = await new Promise<boolean>(resolve =>
+        session.protocol.isProtocolHandled(SingleSignOn.SSO_PROTOCOL, result => resolve(result as any)),
```
ffflorian: The types state that the callback for `isProtocolHandled` returns an error but it actually returns a boolean.

PR #3166, opened 2019-10-29
chore: Add CircleCI integration, remove GitHub actions

ffflorian:


.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  build_test_windows:
+    executor: win/vs2019
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
```
bennyn: Is the fallback necessary?

ffflorian: Not sure, I took this from the example config.

PR #3197, opened 2019-11-01
refactor: Add more logging and focused window check

ffflorian:


electron/src/lib/CoreProtocol.ts
```diff
@@ -85,12 +98,16 @@ export class CustomProtocolHandler {
       });
     } else {
       app.once('ready', async () => {
-        const url = process.argv[CORE_PROTOCOL_POSITION];
```
ffflorian: Until now it was just assumed that the protocol link would be the first argument. In some cases this was wrong so now we are just looking through the arguments where it is.

PR #3246, opened 2019-11-11
runfix: Re-enable global shortcuts for macOS (WEBAPP-6528)

ffflorian:
This re-enables global shortcuts for macOS since the trick from https://github.com/wireapp/wire-desktop/pull/3207 doesn't work here.

### DIFF
--- a/electron/src/menu/system.ts
+++ b/electron/src/menu/system.ts
@@ -418,7 +418,7 @@ export const createMenu = (isFullScreen: boolean): Menu => {
   }
 
   if (Array.isArray(windowTemplate.submenu)) {
-    const muteAccelerator = 'CmdOrCtrl+Alt+M';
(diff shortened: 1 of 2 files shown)


electron/src/menu/system.ts
```diff
@@ -429,13 +429,14 @@ export const createMenu = (isFullScreen: boolean): Menu => {
       visible: false,
     };
 
-    const switchShortcuts: MenuItemConstructorOptions[] = [...Array(config.maximumAccounts).keys()].map(index => {
+    const switchShortcuts: MenuItemConstructorOptions[] = [...Array(config.maximumAccounts)].map((_, index) => {
```
AndyLnd:
```suggestion
    const switchShortcuts: MenuItemConstructorOptions[] = Array.from({length: config.maximumAccounts}, (_, index) => {
```

PR #3329, opened 2019-12-05
refactor: Use optional chaining

ffflorian:


bin/build-tools/lib/build-macos.ts
```diff
@@ -102,7 +102,7 @@ export async function buildMacOSConfig(
       };
     }
 
-    if (macOSConfig.notarizeAppleId && macOSConfig.notarizeApplePassword) {
```
Yserz: This doesn't look correct

ffflorian: Whoops thanks

electron/src/lib/openGraph.ts
```diff
@@ -200,7 +200,7 @@ export const getOpenGraphDataAsync = async (url: string): Promise<OpenGraphResul
     metadata.image = metadata.image[0];
   }
 
-  if (typeof metadata.image === 'object' && !Array.isArray(metadata.image) && metadata.image.url) {
+  if (typeof metadata.image === 'object' && metadata.image?.url) {
```
Yserz: If I see that correctly `metadata.image` can be an object or an array. if it's an array typeof === 'object' will return true. So in case `metadata.image` is an array this will break imho.

ffflorian: AFAIC this will not break since JS will realize that `metadata.image.url` doesn't exist.

babel.config.js
```diff
@@ -1,4 +1,8 @@
-const plugins = ['@babel/plugin-proposal-class-properties', '@babel/plugin-proposal-object-rest-spread'];
+const plugins = [
+  '@babel/plugin-proposal-class-properties',
+  '@babel/plugin-proposal-object-rest-spread',
```
Yserz: We might be able to remove that depending on the Electron version we use

electron/src/lib/openGraph.ts
Yserz: Mh, you are right

babel.config.js
ffflorian: https://github.com/wireapp/wire-desktop/pull/3330/files

PR #3345, opened 2019-12-11
feat: Migrate TSLint to ESLint

lipis:


.eslintrc.json
```diff
@@ -3,10 +3,25 @@
...
   "plugins": ["jasmine"],
   "rules": {
-    "handle-callback-err": 2,
-    "import/no-default-export": 0,
-    "no-undef": 0
+    "import/no-default-export": "off",
+    "no-magic-numbers": "off",
+    "no-undef": "off",
+    "no-unsanitized/property": "off",
+    "simple-import-sort/sort": "off"
```
lipis: We should try to remove some of the `offs` before migrating.

bin/bin-utils.ts
```diff
@@ -51,7 +51,7 @@ export async function restoreFiles({originalPaths, backupPaths, tempDir}: Backup
 }
 
 export const getLogger = (namespace: string, name: string) =>
-  LogFactory.getLogger(name, {namespace: `@wireapp/${namespace}`, forceEnable: true, separator: '/'});
+  LogFactory.getLogger(name, {forceEnable: true, namespace: `@wireapp/${namespace}`, separator: '/'});
```
ffflorian: Sorting :heart_eyes:

electron/src/lib/openGraph.test.main.ts
```diff
@@ -28,8 +28,35 @@ const defaultMessageUtf8 = [72, 101, 108, 108, 111, 32, 102, 114, 111, 109, 32,
...
+  208,
+  189,
+  208,
+  190,
+  208,
+  186,
+  208,
+  176,
+  33,
+];
```
ffflorian: Let's disable ESLint for this line, it's not helpful here.

electron/src/lib/openGraph.test.main.ts
```diff
@@ -103,6 +130,6 @@ describe('openGraph', () => {
   it('saves cookies on requests', async () => {
     const cookieText = 'my-cookie';
     const result = await cookieRequest(cookieText);
-    assert.strictEqual(result.config.headers['Cookie'], cookieText);
+    assert.strictEqual(result.config.headers.Cookie, cookieText);
```
ffflorian: This won't work with TypeScript as `Cookie` is private and should not be accessed outside of tests.

electron/src/menu/TrayHandler.test.main.ts
```diff
@@ -32,11 +32,11 @@ describe('initTray', () => {
...
-    assert.strictEqual(Object.keys(tray['icons']!).length, 3);
-    assert.strictEqual(tray['icons']!.badge.constructor.name, 'NativeImage');
-    assert.strictEqual(tray['icons']!.tray.constructor.name, 'NativeImage');
-    assert.strictEqual(tray['icons']!.trayWithBadge.constructor.name, 'NativeImage');
-    sinon.assert.match(tray['trayIcon']!, sinon.match.defined);
+    assert.strictEqual(Object.keys((tray as any).icons!).length, 3);
+    assert.strictEqual((tray as any).icons!.badge.constructor.name, 'NativeImage');
+    assert.strictEqual((tray as any).icons!.tray.constructor.name, 'NativeImage');
+    assert.strictEqual((tray as any).icons!.trayWithBadge.constructor.name, 'NativeImage');
+    sinon.assert.match((tray as any).trayIcon!, sinon.match.defined);
```
ffflorian: Can we get rid of `any` here and stick to the old format?

electron/src/renderer/preload-sso.ts
```diff
@@ -22,7 +22,6 @@ const {SingleSignOn} = remote.require('./sso/SingleSignOn');
 
 // Only execute the helper if the origin is the backend
 if (typeof document.location?.origin === 'string' && SingleSignOn.isBackendOrigin(document.location.origin)) {
-  // tslint:disable-next-line:no-floating-promises
```
ffflorian: Is there an equivalent in ESLint? If not, we should document that.

electron/src/sso/SingleSignOn.ts
```diff
@@ -19,21 +19,20 @@
...
   session,
+  WebContents,
 } from 'electron';
 import * as path from 'path';
 import {URL} from 'url';
-const minimist = require('minimist');
-
 import {getLogger} from '../logging/getLogger';
 import {config} from '../settings/config';
+const minimist = require('minimist');
```
ffflorian: This should stay at the top with the other library imports.

electron/src/menu/TrayHandler.test.main.ts
lipis: Maybe we should remove the this rule then: https://github.com/wireapp/wire-web-packages/blob/master/packages/eslint-config/eslintrc.json#L19 (`dot-notation`)

lipis: /cc @bennyn @AndyLnd? What do you guys think?

electron/src/lib/openGraph.test.main.ts
lipis: `headers` is `any` in that particular case.. so there are no errors.

.eslintrc.json
```diff
@@ -3,10 +3,25 @@
...
+  ],
+  "parserOptions": {
+    "project": ["./tsconfig.json", "./tsconfig.bin.json", "./tsconfig.mocha.json"]
+  },
   "plugins": ["jasmine"],
   "rules": {
-    "handle-callback-err": 2,
-    "import/no-default-export": 0,
-    "no-undef": 0
+    "import/no-default-export": "off",
```
ffflorian: This needs to stay since it makes more sense to export React components as default.

.eslintrc.json
```diff
@@ -3,10 +3,25 @@
...
+    "no-magic-numbers": "off",
```
ffflorian: This needs to stay since numbers which are declared as static members with a name are also called "magic numbers", which doesn't make sense IMO.

.eslintrc.json
```diff
@@ -3,10 +3,25 @@
...
+    "no-undef": "off",
```
ffflorian: This needs to stay since ESLint has some problems with `import * as <var>`, complaining that `<var>` would be undefined.

.eslintrc.json
ffflorian: See my comments

.eslintrc.json
lipis: I agree... this doesn't make sense..

.eslintrc.json
lipis: Shall we remove that form the package then for all projects?! Most of the project will have some kind of React in it..

ffflorian: Mh no, I would like to find out first if we can somehow tell ESLint to disable some rules for certain directories...

PR #3482, opened 2020-01-20
fix: Use single instance lock instead of variable

ffflorian:


bennyn:
It broke the macOS build. 🦀

PR #3486, opened 2020-01-21
feat: Receive custom WebApp URL

ffflorian:


electron/src/renderer/preload-webview.ts
```diff
@@ -94,14 +94,27 @@ const subscribeToWebappEvents = () => {
   window.amplify.subscribe(window.z.event.WebApp.TEAM.INFO, (info: TeamAccountInfo) => {
     const debugInfo = {
       ...info,
-      picture: typeof info.picture === 'string' ? `${info.picture.substring(0, 100)}...` : '',
+      picture: `${String(info.picture).substring(0, 100)}...`,
```
bennyn: When `info.picture` is `undefined` then we will get `undefined...` instead of `''`. This is a behaviour change which I will revert because it seems unwanted.

PR #3512, opened 2020-01-27
feat: Clear cache on exit

lipis:


electron/src/runtime/lifecycle.ts
```diff
@@ -69,15 +69,17 @@ export const getWebViewId = (contents: WebContents): string | undefined => {
 // Using exit instead of quit for the time being
 // see: https://github.com/electron/electron/issues/8862#issuecomment-294303518
 export const quit = (): void => {
-  logger.info('Initiating app quit ...');
+  logger.info('Initiating app quit...');
   settings.persistToFile();
+  logger.info('Clear cache...');
+  session.defaultSession.clearCache().catch(() => {});
```
lipis: that's the actual change!

bennyn: Ok, then let's keep this PR to that change and put back the spaces to "...". These were set on purpose.

bennyn:
```suggestion
  session.defaultSession.clearCache().catch(() => {});
```

There should be at least an error log when catching. Just swallowing the error is not a good practice.

lipis: From the copy point of view we should never have spaces before `,.?! etc`.

Shall we fix that in separate PR?

electron/src/runtime/lifecycle.ts
```diff
@@ -71,6 +71,10 @@ export const getWebViewId = (contents: WebContents): string | undefined => {
 export const quit = (): void => {
   logger.info('Initiating app quit ...');
   settings.persistToFile();
+  logger.info('Clear cache ...');
+  session.defaultSession.clearCache().catch(error => {
+    logger.info(error);
```
bennyn: Let's flag it as an error or at least a warning.

electron/src/runtime/lifecycle.ts
bennyn: No. As I wrote. The spaces were set on purpose. They should be kept.

electron/src/runtime/lifecycle.ts
lipis: done..

PR #3547, opened 2020-02-04
feat: Allow navigation for custom URL on account level

bennyn:


electron/src/renderer/preload-webview.ts
```diff
@@ -103,22 +103,16 @@ const subscribeToWebappEvents = () => {
...
-          accountId,
-          customURL: data.url,
-        });
-      }
-    },
-    {
-      once: true,
-    },
-  );
+  window.addEventListener('NavigationEvent', event => {
```
Yserz: NavigationEvent could be an enum

electron/src/renderer/preload-webview.ts
```diff
@@ -103,22 +103,16 @@ const subscribeToWebappEvents = () => {
...
+    const accountId = new URLSearchParams(window.location.search).get('id');
+    const data = (event as CustomEvent).detail;
+    if (data) {
+      ipcRenderer.send(EVENT_TYPE.WRAPPER.CUSTOM_WEBAPP, {
```
Yserz: Would be nice if the naming of the internal event would match the name of the external event (NavigationEvent)

electron/src/window/WindowUrl.ts
```diff
@@ -16,17 +16,22 @@ export class WindowUrl {
...
-
     const newEnvUrl = WindowUrl.replaceQueryParams(customBackendUrl, mergedParams);
 
-    const cutPattern = '?env=';
-    const localFilePrefix = localRendererUrl.substring(0, localRendererUrl.indexOf(cutPattern) + cutPattern.length);
-
-    return `${localFilePrefix}${newEnvUrl}`;
+    if (withRendererPage === true) {
+      const cutPattern = '?env=';
+      const localFilePrefix = localRendererUrl.substring(0, localRendererUrl.indexOf(cutPattern) + cutPattern.length);
```
Yserz:
```suggestion
      const localFilePrefix = localRendererUrl.split(cutPattern)[0];
```

electron/src/main.ts
```diff
@@ -223,7 +207,7 @@ const showMainWindow = async (mainWindowState: WindowStateKeeper.State) => {
     webPreferences: {
       backgroundThrottling: false,
       enableBlinkFeatures: '',
-      nodeIntegration: false,
+      nodeIntegration: true,
```
Yserz: As discussed that's something we will resolve in a future PR

PR #3605, opened 2020-02-24
feat: Local web server integration

pwnsdx:


electron/src/renderer/preload-app.ts
```diff
@@ -99,7 +99,7 @@ const setupIpcInterface = (): void => {
       }
 
       console.info(`Processing deletion of "${accountID}"`);
-      const viewInstanceId = accountWebview.getWebContents?.().id;
+      const viewInstanceId = accountWebview.getWebContentsId();
```
Yserz: I think that breaks when deleting an account while being on the error page (no webview is rendered)

electron/wire.json
```diff
@@ -18,7 +18,7 @@
   "privacyUrl": "https://wire.com/privacy/",
   "raygunApiKey": "",
   "supportUrl": "https://support.wire.com",
-  "version": "3.14.0",
+  "version": "3.13.0",
```
Yserz: 3.14 would be the correct version atm

PR #3641, opened 2020-03-05
runfix: Fix issues with Windows and Linux deployment

mythsunwind:


jenkins/deployment.groovy
```diff
@@ -76,120 +76,95 @@ node('master') {
...
-            }
-          } else {
-            withCredentials([
-              string(credentialsId: 'MACOS_HOCKEY_ID', variable: 'MACOS_HOCKEY_ID'),
-              string(credentialsId: 'MACOS_HOCKEY_TOKEN', variable: 'MACOS_HOCKEY_TOKEN')
-            ]) {
-              MACOS_HOCKEY_ID = env.MACOS_HOCKEY_ID
-              MACOS_HOCKEY_TOKEN = env.MACOS_HOCKEY_TOKEN
-            }
+            appName = 'Wire by XY - macOS'
```
bennyn: Does this need to be a valid App Center key?

mythsunwind: I now throw an error (the script will stop directly) so that we do not forget to add real values

mythsunwind: Please re-check again

PR #3657, opened 2020-03-10
fix: Only show error page for requests to current webapp origin

Yserz:


electron/renderer/src/components/Webview.jsx
```diff
@@ -104,15 +104,21 @@ const Webview = ({
   }, [account]);
 
   useEffect(() => {
-    const listener = error => setWebviewError(error);
+    const listener = error => {
+      const urlOrigin = new URL(getEnvironmentUrl(account)).origin;
+      console.warn(`WEBVIEW fired did-fail-load for URL "${error.validatedURL}" and account ID "${account.id}"`);
```
bennyn:
```suggestion
      console.warn(`WEBVIEW fired "did-fail-load" for URL "${error.validatedURL}" and account ID "${account.id}"`);
```

PR #3711, opened 2020-03-23
fix: Check proxy protocols explicitly

Yserz:


electron/src/main.ts
```diff
@@ -104,6 +104,9 @@ if (argv.version) {
 if (argv[ARG.PROXY_ARG]) {
   try {
     proxyInfoArg = new URL(argv[ARG.PROXY_ARG]);
+    if (proxyInfoArg.protocol !== 'http:' && proxyInfoArg.protocol !== 'https:') {
```
ffflorian: What about `socks5://`? :)

Yserz: Done

electron/src/main.ts
```diff
@@ -104,6 +104,14 @@ if (argv.version) {
 if (argv[ARG.PROXY_ARG]) {
   try {
     proxyInfoArg = new URL(argv[ARG.PROXY_ARG]);
+    if (
+      proxyInfoArg.protocol !== 'http:' &&
```
bennyn: Can this be done: `if (!['http:', 'https:', 'socks4:', 'socks5:'].includes(proxyInfoArg.protocol))`?

ffflorian: Or:
```js
if (!/^(https?|socks[45]):/.test(proxyInfoArg.protocol))
```

PR #3834, opened 2020-04-30
fix: Log web console messages to file (WEBAPP-6784)

ffflorian:


electron/src/main.ts
```diff
@@ -544,11 +544,10 @@ class ElectronWrapperInit {
             contents.on('console-message', async (_event, _level, message) => {
               const webViewId = lifecycle.getWebViewId(contents);
               if (webViewId) {
-                const logFilePath = path.join(LOG_DIR, webViewId, config.logFileName);
                 try {
-                  await LogFactory.writeMessage(message, logFilePath);
+                  await LogFactory.writeMessage(message.replace(/%c(.+)%c /, '$1 '), LOG_FILE);
```
bennyn, line 548: Please put the regex `/%c(.+)%c` into a variable which gives information that it is about the color code.

electron/src/main.ts
```diff
@@ -363,7 +363,7 @@ const handleAppEvents = () => {
             logger.log('Proxy prompt was submitted');
 
             const {username, password} = promptData;
-            const protocol: string | undefined = proxyInfoArg?.protocol?.replace(':', '');
+            const protocol = proxyInfoArg?.protocol?.replace(':', '');
```
bennyn, line 366: What's the intention here? Can you comment it or have a test for it?

ffflorian: The intention was to remove useless TypeScript annotations.

bennyn: I don't think that the intention of `replace(':', '')` is to remove useless TypeScript annotations or is it?

ffflorian: What do you mean? This line was added [in #3634](https://github.com/wireapp/wire-desktop/commit/dda822ed31fe636a4a70205fad898430a2ba9ee0#diff-acab454f418662fe6f697507e4c4103aR367), I just removed the type annotations.

PR #3924, opened 2020-05-26
fix: Verify certificate validated by Chromium (WEBAPP-6864)

bennyn:


franziskuskiefer:
@bennyn I don't see tests in here. Where can I find them? Or if there aren't any, where's the issue to implement them?

PR #3975, opened 2020-06-16
fix: Recommend filename without : in the timestamp

lipis:


electron/src/lib/download.ts
```diff
@@ -71,3 +69,9 @@ export const downloadFile = async (bytes: Uint8Array, filename: string, options?
     logger.error(error);
   }
 };
+
+export const suggestedFileName = (timestamp?: string) => {
```
bennyn, line 73:
```suggestion
export const suggestFileName = (timestamp?: string) => {
```

electron/src/lib/download.test.main.ts
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import * as assert from 'assert';
+import {suggestedFileName} from './download';
+
+describe('download', () => {
+  it('suggests the correct filename', async () => {
```
bennyn, line 24:
```suggestion
  it('converts colons to dashes because colons cannot be used in filenames on Windows', async () => {
```

PR #4015, opened 2020-07-02
feat: Add native spellchecker support (WEBAPP-6583)

ffflorian:


electron/src/renderer/menu/context.ts
```diff
@@ -74,63 +114,38 @@ const imageMenu: ElectronMenuWithImageAndTime = Menu.buildFromTemplate([
...
+  } else if (params.mediaType === 'image') {
+    imageMenu.image = params.srcURL;
+    imageMenu.popup({window});
+  } else if (!!params.linkURL) {
+    const copyContext = params.linkURL.replace(/^mailto:/, '');
+    createDefaultMenu(copyContext).popup({window});
+  } else if (!!params.selectionText) {
+    const copyContext = params.selectionText;
+    createDefaultMenu(copyContext).popup({window});
+  } else if (params.editFlags.canCopy) {
```
bennyn, line 132: This condition does the same as `!!params.selectionText`. Can we put them in one if block connected with a or condition?

PR #4040, opened 2020-07-07
fix: Remove redundant configuration saving

ffflorian:


electron/src/main.ts
```diff
@@ -341,7 +341,6 @@ const handleAppEvents = () => {
   });
 
   app.on('before-quit', () => {
-    settings.persistToFile();
```
ffflorian, line 344: This is handled by `lifecycle.quit()`

PR #4045, opened 2020-07-08
feat: Show the proxy prompt again if the credentials are invalid (WEBAPP-4574)

ffflorian:


JamesDBartlett:
Something broke with this update. I launch Wire from a shortcut to the Wire app executable that passes the command line arguments I need, like this:
`C:\path\to\wire\Wire.exe --proxy-server="https://username:password@192.168.1.2:1080"`

Since installing this update, Wire no longer accepts the basic authentication, and forces me to enter the username and password via the dialog every time I run the app. I've been using the same shortcut for months, and haven't changed anything on the proxy server either. Other applications using the same proxy are still working without issue.

PR #4112, opened 2020-08-07
fix: Do not flash application window when user status is busy (webapp-7046)

thisisamir98:


electron/src/main.ts
```diff
@@ -147,8 +147,8 @@ const bindIpcEvents = (): void => {
     WindowManager.showPrimaryWindow();
   });
 
-  ipcMain.on(EVENT_TYPE.UI.BADGE_COUNT, (_event, count: number) => {
-    tray.showUnreadCount(main, count);
+  ipcMain.on(EVENT_TYPE.UI.BADGE_COUNT, (_event, {count, ignoreFlash}: {count: number; ignoreFlash: boolean}) => {
```
ffflorian, line 150: `count` and `ignoreFlash` are optional

thisisamir98: done in https://github.com/wireapp/wire-desktop/pull/4112/commits/609b6180bde42c669d4a85a92a2c2ad1ac6d864a

PR #4159, opened 2020-08-17
fix: Don't allow reusing the renderer process

ffflorian:


electron/src/main.ts
```diff
@@ -137,6 +137,10 @@ Object.entries(config).forEach(([key, value]) => {
 // Squirrel setup
 app.setAppUserModelId(`com.squirrel.wire.${config.name.toLowerCase()}`);
 
+// TODO: Use a Context Aware Plugin for Link Previews,
```
bennyn, line 140: Instead of a TODO would be good mentioning here why we need to deactivate `allowRendererProcessReuse`.

PR #4247, opened 2020-09-08
runfix: Set maximum zoom limits and fix zoom in shortcut

ffflorian:


electron/src/menu/system.ts
```diff
@@ -241,7 +241,7 @@ const windowTemplate: MenuItemConstructorOptions = {
       label: locale.getText('menuActualSize'),
     },
     {
-      accelerator: 'CmdOrCtrl+Plus',
+      accelerator: 'CmdOrCtrl+=',
```
ffflorian, line 244: Seems like <kbd>Ctrl</kbd>+<kbd>Plus</kbd> does not actually trigger <kbd>+</kbd> while <kbd>Ctrl</kbd>+<kbd>=</kbd> does.

See https://github.com/electron/electron/issues/1507#issuecomment-118424331.

bennyn: Ok, please add this info (issue link) to the code itself.

bennyn (changes requested):

Yserz (approved):

PR #4293, opened 2020-09-17
fix: Add times for Windows update check log

ffflorian:


electron/src/update/squirrel.ts
```diff
@@ -129,14 +129,16 @@ export async function installUpdate(): Promise<void> {
...
-  const nextCheck = config.squirrelUpdateInterval.DELAY;
-  const everyCheck = config.squirrelUpdateInterval.INTERVAL;
-  const readableNextCheck = StringUtil.pluralize('minute', nextCheck / MINUTE_IN_MILLIS);
-  const readableEveryCheck = StringUtil.pluralize('hour', everyCheck / HOUR_IN_MILLIS);
-  logger.info(`Scheduling Windows update to check in "${readableNextCheck}" and every "${readableEveryCheck}" ...`);
-
-  setTimeout(installUpdate, nextCheck);
-  setInterval(installUpdate, everyCheck);
+  const squirrelDelay = config.squirrelUpdateInterval.DELAY;
+  const squirrelInterval = config.squirrelUpdateInterval.INTERVAL;
```
lipis, line 133: Not sure if we really save much with these two const: `squirrelDelay`, `squirrelInterval`

lipis (approved):

electron/src/update/squirrel.ts
ffflorian: We are using it twice, so it was reason enough for me to make a constant out of it. :slightly_smiling_face:

PR #4408, opened 2020-10-19
fix(deploy-tools): Log upload error correctly

ffflorian:


bin/deploy-tools/lib/GitHubDraftDeployer.ts
```diff
@@ -166,23 +166,32 @@ export class GitHubDraftDeployer {
     try {
       await axios.post(url, file, {headers, maxContentLength: TWO_HUNDRED_MB_IN_BYTES});
     } catch (uploadError) {
-      this.logger.error('Error response from GitHub:', uploadError.response.data);
```
ffflorian, line 169: In this case `response` was `undefined`

lipis (approved):

PR #4591, opened 2020-12-08
feat: Open DevTools by account index (SQCORE-283)

ffflorian:
The DevTools can now be opened as follows:

* `--devtools` / `--devtools=0`: Opens the renderer's DevTools
* `--devtools=1`: Opens the first account's DevTools
* `--devtools=2`: Opens the second account's DevTools
... etc.


electron/src/main.ts
```diff
@@ -235,8 +235,14 @@ const showMainWindow = async (mainWindowState: windowStateKeeper.State): Promise
     webappURL.hash = customProtocolHandler.hashLocation;
   }
 
-  if (argv.devtools) {
-    main.webContents.openDevTools({mode: 'detach'});
+  if (typeof argv.devtools !== 'undefined') {
+    if (typeof argv.devtools === 'number') {
```
bennycode, line 239: If the type is a number then it is also undefined, so we wouldn't need two if cases or am I missing something?

electron/src/menu/developer.ts
```diff
@@ -37,12 +37,16 @@ const reloadTemplate: MenuItemConstructorOptions = {
...
 
-const openDevTools = async (webViewIndex: number): Promise<void> => {
-  const snippet = `document.getElementsByTagName("webview")[${webViewIndex}].openDevTools({mode: "detach"})`;
-
+export const openDevTools = async (webViewIndex?: number): Promise<void> => {
   const primaryWindow = WindowManager.getPrimaryWindow();
+
   if (primaryWindow) {
-    await executeJavaScriptWithoutResult(snippet, primaryWindow.webContents);
+    if (typeof webViewIndex === 'number' && webViewIndex > 0) {
```
bennycode, line 44:
```suggestion
    if (webViewIndex > 0) {
```

ffflorian: `Object is possibly 'undefined' .ts(2532)`

bennycode (approved):

PR #4841, opened 2021-03-03
runfix: Use opacity instead of TransitionContainer on the LoadingSpinner

ffflorian:
Since we removed only the `FlexBox` and the Loading Spinner itself after the page is loaded, we had a left-over element over the whole interface which rendered the app useless :slightly_smiling_face:

### DIFF
--- a/electron/renderer/src/components/LoadingSpinner.css
+++ b/electron/renderer/src/components/LoadingSpinner.css
@@ -1,0 +1,33 @@
+/*
+ * Wire
+ * Copyright (C) 2021 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+#loading-spinner-wrapper {
+  display: flex;
+  height: 100vh;
+  position: absolute;
+  width: 100%;
+  z-index: 99;
+  transition: opacity 0.25s ease-in-out;
+  -moz-transition: opacity 0.25s ease-in-out;
+  -webkit-transition: opacity 0.25s ease-in-out;
+}
+
+#loading-spinner-box {
+  margin: 0 auto;
+}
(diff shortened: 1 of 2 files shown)


electron/renderer/src/components/LoadingSpinner.css
```diff
@@ -0,0 +1,33 @@
...
+
+#loading-spinner-wrapper {
+  display: flex;
+  height: 100vh;
+  position: absolute;
+  width: 100%;
+  z-index: 99;
+  transition: opacity 0.25s ease-in-out;
+  -moz-transition: opacity 0.25s ease-in-out;
+  -webkit-transition: opacity 0.25s ease-in-out;
```
thisisamir98, line 28: does not compiler add these automatically?

electron/renderer/src/components/LoadingSpinner.css
```diff
@@ -0,0 +1,33 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+#loading-spinner-wrapper {
```
thisisamir98, line 20: if you're going to use this multiple times in a page I suggested to not use an ID and use classes instead but if it's just one place it's fine...

thisisamir98 (approved):

electron/renderer/src/components/LoadingSpinner.css
ffflorian: What compiler? I am using css :)

electron/renderer/src/components/LoadingSpinner.css
ffflorian: Mh you are right, we may be using it in multiple webviews..

electron/renderer/src/components/LoadingSpinner.css
Yserz: Bardia means something like postCSS as a CSS pre-compiler.

ffflorian: Ah okay, we don't have that in wire-desktop.

electron/renderer/src/components/LoadingSpinner.css
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+.loading-spinner-wrapper {
+  display: flex;
+  height: 100vh;
+  position: absolute;
+  width: 100%;
+  z-index: 99;
+  transition: opacity 0.25s ease-in-out;
+  -moz-transition: opacity 0.25s ease-in-out;
```
bennycode, line 27: Wire desktop runs a Chromium, so we don't need the `-moz-` prefix, or am I wrong?

ffflorian: True!

Issue #2507, opened 2019-04-15
Wire Web/Desktop constantly uses CPU during idle


rawlife56:
Wire version: 3.9.2895 <Check the About dialog>
Wire for web version: 2019.04.10.0901<Check the About dialog>
Operating system: Linux (Arch-KDE) 

What steps will reproduce the problem?

1. Open Wire
2. Close and let it sit in the tray
3. Check the constant CPU usage of 2-4% with some spikes upto 10% in regular intervals without any incoming messages.

What is the expected result?
No CPU usage and minimal CPU usage in intervals if needed. 

What is the actual result?
Constant CPU usage of 2-4% with some spikes upto 10% in regular interval

Workarounds I tried-
Searched across issues and found few like [these](https://github.com/wireapp/wire-desktop/issues/135) which are quite old & irrelevant. 

Strange findings- 
1. It also happens in the web app. Tried in Firefox browser and [Rambox](https://github.com/TheGoddessInari/rambox) with the clean chats, both resulted in the same constant CPU usage. 
2. Few specific conversations result in constant 10% CPU usage. (including idle.)

_A note about screenshots_ - Every Electron instance running in this pic is from Wire. This doesn't happen with other electron applications. I took extreme care before capturing these screenshots verifying it's only the Wire running and also did check whether killing wire would terminate these processes too. Yes, they did terminate if I kill wire. 

![Screenshot_20190416_015407](https://user-images.githubusercontent.com/42450674/56164702-e1ffbe80-5fee-11e9-91a2-7c3b57e4c3c9.png)'
![Screenshot_20190416_021201](https://user-images.githubusercontent.com/42450674/56164719-edeb8080-5fee-11e9-99f0-88c37465694c.png)

fermulator:
This is a regression I think, I am observing new significant degrade in battery life on notebook for simpler reasons.

```
 ps wauxxx | grep wire-desktop | sort -rnk3 | head -1
[...]

Issue #2239, opened 2019-02-21
Poor performance on Linux


hmenke:
Wire version: 3.6.2885 (latest from `deb https://wire-app.wire.com/linux/debian stable main`)
Wire for web version: -
Operating system: Debian GNU/Linux 9.8 (stretch) Linux 4.19.0-0.bpo.2-amd64
Which antivirus software do you have installed: -

### What steps will reproduce the problem?

1.  Start `wire-desktop`
2.  Make a video call

### What is the expected result?

Reasonable CPU and RAM usage.

### What is the actual result?

Enormous CPU and RAM usage to an extent where the system becomes unresponsive and the frame rate of the video call drops to about 2 FPS.

### Please provide any additional information below.

Screenshot of `htop` while I'm in the video call.

![test](https://user-images.githubusercontent.com/1903556/53156828-1be8b180-35b8-11e9-8951-81242ca0ef59.png)

hmenke:
#661 might be related.  I'll try the fixes proposed there.

hmenke:
I have applied the fix proposed in #661, namely to define the environment variable `ELECTRON_FORCE_WINDOW_MENU_BAR`.  This improved the performance insofar that Wire now “only” uses 90% instead of 100% on all cores.  The memory consumption was also improved a little and the rest of system is no longer unusable while I'm talking on Wire.  Unfortunately, the Wire application itself remains barely usable.  After entering the call the video frame rate drops to about 2 FPS within the first minute.  The whole interface is reacting slowly.

![test](https://user-images.githubusercontent.com/1903556/53226583-6e3ad880-3673-11e9-8e20-30c5faafec61.png)

ffflorian:
Hi @hmenke, do I see correctly that you have 4 GB of memory?

hmenke:
@ffflorian Yes, I have 4GB of physical memory and 10GB of swap.  However, memory doesn't seem to be the problem, especially after applying the fix from #661 which improved idle performance of Wire *a lot!*  Is there a way to profile Wire to see where it is spending the most time?  Is there something like a compiler switch like `-pg` for GCC to generate instrumentation for `gprof`?

ffflorian:
> Is there a way to profile Wire to see where it is spending the most time? Is there something like a compiler switch like [`-pg` for GCC](https://gcc.gnu.org/onlinedocs/gcc-8.3.0/gcc/Instrumentation-Options.html) to generate instrumentation for `gprof`?

[...]

PR #5164, opened 2021-07-06
fix: Don't render LoadingSpinner after it has faded out

jviide:
This pull request attempts to reduce Wire's CPU usage when idle.

In the discussions related to issues #2507 and #2239 (and maybe elsewhere) @charlag et al. tracked down a culprit for some surpsingly high constant CPU load. The LoadingSpinner component is always rendered on top of the rest of the UI with 0 opacity after it has initially faded out, continuously using resources.

This pull request addresses the issue ~~by hooking to the `onTransitionEnd` and `onTransitionCanceled` events and set a flag marking the fade out completed~~ giving the fade out transition a fixed amount of time to finish. After this the component returns `null`, effectively removing the LoadingSpinner contents from the DOM tree.

Personally, I've noticed constant ~15% load when Wire was idling on my M1 MacBook Air. Launching Wire with devtools enabled and removing the spinner from the DOM drops the idle CPU usage to near zero. I couldn't get the build infra work 100% on my machine, so @ikisusi and @esatormi volunteered to test this patch on their Intel Macs, both getting similar results. @ikisusi's CPU activity monitor looked like this _before_ applying the patch (and after the spinned had faded to 0 opacity):

before

And like this _after_ applying the patch:

after

### DIFF
--- a/electron/renderer/src/components/LoadingSpinner.css
+++ b/electron/renderer/src/components/LoadingSpinner.css
@@ -19,6 +19,7 @@
 
 .loading-spinner-wrapper {
   height: 100vh;
+  display: flex;
   position: absolute;
   width: 100%;
   z-index: 2;
--- a/electron/renderer/src/components/LoadingSpinner.jsx
+++ b/electron/renderer/src/components/LoadingSpinner.jsx
@@ -22,8 +22,11 @@ import {FlexBox, Loading, COLOR} from '@wireapp/react-ui-kit';
 
 import './LoadingSpinner.css';
 
+const TRANSITION_GRACE_PERIOD_MS = 500;
+
 const LoadingSpinner = ({visible, webviewRef}) => {
   const [isLoading, setIsLoading] = useState(true);
+  const [isFinished, setIsFinished] = useState(false);
 
   useEffect(() => {
     const webview = webviewRef.current;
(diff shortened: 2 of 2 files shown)


ffflorian:
Hi @jviide, thanks for the PR! I'll review it tomorrow.

jviide:
I modified the approach. Instead of observing `onTransitionEnd` and `onTransitionCancel` events, the transition is given a fixed amount of time to finish after the `isLoading` flag flips from true to false. This happens regardless of whether the `visible` flag is on or off.

The length of this "transition grace period" is 500 ms by default to accommodate the current 250 ms transition.

I made this change after realizing that there may be situations when the transition events never get dispatched. For example, the user might switch accounts just right, making visible a webview that has already finished loading in the background. Then the transition from opacity 1 to 0 never happens.

Yserz (approved):

ffflorian:
Tested manually. Looks good!

ffflorian (approved):

ffflorian:
Again thank you very much @jviide!

bennycode:
Hi @jviide, your fix will go live with v3.27. We plan to ship this release early next week.

bennycode:
Hi @marvinhagemeister, unfortunately our release got pushed back so it comes later than I anticipated. I am also not working at Wire anymore so my successor will take care of it.

fluffypony:
Still waiting for this to go in, and for M1 support (Electron has had M1 support for [well over a year](https://www.electronjs.org/blog/apple-silicon)).

JMoVS:
@Yserz Is there any news when this might finally land on desktop on Macs? Desktop is still stuck at 3.26

fluffypony:
This is crazy that it's taking months to go in - is this not something that can be part of a Wire Web update? I was under the impression [those happen weekly](https://support.wire.com/hc/en-us/articles/115001919905), yet here we are nearly half a year later.

embeddedt:
Can confirm that this is still an issue in January 2022 - after using the [uBlock-based workaround](https://github.com/wireapp/wire-desktop/issues/2507#issuecomment-833607881) suggested here, my CPU usage at idle is a lot less.

jviide:
Seems that a version containing this PR has now been released to Mac App Store.

PR #5549, opened 2022-02-02
feat: Upgrade to Electron 13

thisisamir98:


electron/src/main.ts
```diff
@@ -117,7 +118,7 @@ if (argv[config.ARGUMENT.PROXY_SERVER] || fileBasedProxyConfig) {
       throw new Error('No protocol for the proxy server specified.');
     }
   } catch (error) {
-    logger.error(`Could not parse authenticated proxy URL: "${error.message}"`);
```
tlebon, line 120: do we have some error interface we can use instead of any?

thisisamir98: I did this because of linter errors, have no idea if there is an interface for it

tlebon (approved):

PR #5735, opened 2022-05-24
fix(services): Correctly forward deep links when starting the app (SQSERVICES-1415)

thisisamir98:


electron/src/lib/CoreProtocol.ts
```diff
@@ -134,30 +134,27 @@ export class CustomProtocolHandler {
...
+    app.once('ready', async () => {
+      logger.info('App ready, looking for deep link in arguments ...');
+      const deepLink = this.findDeepLink(process.argv);
+      if (deepLink) {
+        await this.dispatchDeepLink(deepLink);
+      } else {
+        logger.info('No deep link found in arguments.');
+      }
+    });
+    app.on('second-instance', async (_event, argv) => {
```
atomrc, line 150: Since currently this event is not registered for MacOS, Let's maybe prevent registering it for MacOS only. We don't really know what could be the side effect of it. 
WDYT?

electron/src/window/WindowManager.ts
```diff
@@ -68,14 +69,25 @@ export class WindowManager {
...
 
     const primaryWindow = WindowManager.getPrimaryWindow();
 
     if (primaryWindow) {
       if (primaryWindow.webContents.isLoading()) {
-        primaryWindow.webContents.once('did-finish-load', () => primaryWindow.webContents.send(action, ...args));
+        primaryWindow.webContents.once('did-finish-load', () => {
+          WindowManager.actionsQueue = [...WindowManager.actionsQueue, {action, args}];
+          primaryWindow.webContents.send(action, ...args);
+        });
```
atomrc, line 90: We do not need to listen for the `did-finish-load` event anymore, do we? 

```suggestion
// If the webapp is not yet loaded we queue the action we want to send. It will be flushed later on by the flushActionsQueue` method
WindowManager.actionsQueue.push({action, args});
```

electron/src/main.ts
```diff
@@ -171,6 +171,7 @@ const bindIpcEvents = (): void => {
   });
 
   ipcMain.on(EVENT_TYPE.ACTION.NOTIFICATION_CLICK, () => WindowManager.showPrimaryWindow());
+  ipcMain.on(EVENT_TYPE.WEBAPP.APP_LOADED, WindowManager.flushActionsQueue);
```
atomrc, line 174: Point-free can be risky if, in the future, the `flushActionQueue` starts to take an argument. 

These are quite hard bugs to fix

```suggestion
  ipcMain.on(EVENT_TYPE.WEBAPP.APP_LOADED, () => WindowManager.flushActionsQueue());
```

atomrc (changes requested):

electron/src/main.ts
```diff
@@ -171,6 +171,9 @@ const bindIpcEvents = (): void => {
   });
 
   ipcMain.on(EVENT_TYPE.ACTION.NOTIFICATION_CLICK, () => WindowManager.showPrimaryWindow());
+  ipcMain.on(EVENT_TYPE.WEBAPP.APP_LOADED, _event => {
+    WindowManager.flushActionsQueue();
+  });
```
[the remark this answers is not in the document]
atomrc: It's actually not such a good idea. See https://github.com/wireapp/wire-desktop/pull/5735#discussion_r880562561 for context

atomrc (approved):

PR #6069, opened 2022-10-12
chore: keep packages below esm versions

tlebon:
Some packages are updating to esm modules and electron currently does not support esm modules. 

https://github.com/electron/electron/issues/21457


.github/dependabot.yml
```diff
@@ -19,6 +19,16 @@ updates:
...
+      # these are being updated to esm modules after these versions and electron doesn't currently natively support it
+      - dependency-name: image-type
+        versions:
+          - '> 4.x'
+      - dependency-name: globby
+        versions:
+          - '> 11.x'
+      - dependency-name: uuidjs
+        versions:
+          - '> 5.x'
```
atomrc, line 31: I think we want to stop before version 5, isn't it?

atomrc (approved):

PR #6070, opened 2022-10-12
fix: use correct directory to link binaries on deb install

mythsunwind:
Fix issue with post install script after electron-builder update.


tlebon (approved):
L(inux)GTM! 🐧

PR #4965, opened 2021-04-26
fix: Create shortcuts manually to re-enable clickable notifications (SQCORE-544)

ffflorian:
This fixes https://github.com/wireapp/wire-desktop/issues/4160.

PR #6340, opened 2023-01-19
fix(windows): Feed root level exe file to app shortcuts

atomrc:
Since https://github.com/wireapp/wire-desktop/pull/4965 we manually create the shortcuts to the application exec file. 
But because those shortcuts are based on the current `process.execPath` that resolves to the versioned exec file, we get shortcuts that link to a version that might at some point be an old version. 

We now use the root level `Wire.exe` file as the link target

### DIFF
--- a/electron/src/update/squirrel.ts
+++ b/electron/src/update/squirrel.ts
@@ -104,29 +104,32 @@ async function spawnUpdate(args: string[]): Promise<void> {
   }
 }
 
-function createShortcut(location: string): boolean {
+function createShortcut(location: string, execPath: string): boolean {
   // As documented in https://github.com/electron/windows-installer/issues/296,
   // Squirrel has problems with notification clicks on Windows 10.
   // The easiest workaround is to create shortcuts on our own.
-  const shortcutExists = fs.existsSync(location);
+  ///const shortcutExists = fs.existsSync(location);
 
-  return shell.writeShortcutLink(location, shortcutExists ? 'replace' : 'create', {
+  return shell.writeShortcutLink(location, 'replace', {
     appUserModelId: config.appUserModelId,
-    target: process.execPath,
+    target: execPath,
   });
 }
 
 function createShortcuts(): void {
+  childProcess.spawnSync(updateDotExe, ['--createShortcut', `${config.name}.exe`]);
+  const linkPath = fs.readlinkSync(startShortcut);
+
   logger.info('Creating shortcut in the start menu ...');
-  const startResult = createShortcut(startShortcut);
+  const startResult = createShortcut(startShortcut, linkPath);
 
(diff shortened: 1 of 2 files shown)


PatrykBuniX (approved):

electron/src/update/squirrel.ts
```diff
@@ -104,46 +97,17 @@ async function spawnUpdate(args: string[]): Promise<void> {
   }
 }
 
-function createShortcut(location: string): boolean {
-  // As documented in https://github.com/electron/windows-installer/issues/296,
-  // Squirrel has problems with notification clicks on Windows 10.
-  // The easiest workaround is to create shortcuts on our own.
```
V-Gira, line 110: Looks like we're not doing anything to resolve that issue anymore, something changed?

electron/src/update/squirrel.ts
```diff
@@ -104,46 +97,17 @@ async function spawnUpdate(args: string[]): Promise<void> {
...
-  const shortcutExists = fs.existsSync(location);
-
-  return shell.writeShortcutLink(location, shortcutExists ? 'replace' : 'create', {
-    appUserModelId: config.appUserModelId,
-    target: process.execPath,
-  });
+function createShortcuts(): Promise<void> {
+  return spawnUpdate(['--createShortcut', `${config.name}.exe`]);
```
V-Gira, line 101: Ah ok, that's the only thing that's needed, a different .exe name?

electron/src/update/squirrel.ts
atomrc: We did retry to go native and avoid custom code. It seems that the problem is no longer existing... Will give it a few more tries and I'm planning on asking QA to focus on this topic as well :)

V-Gira (approved):

PatrykBuniX (approved):
Makes sense 👌

electron/src/update/squirrel.ts
```diff
@@ -29,20 +29,17 @@ import {getLogger} from '../logging/getLogger';
...
 import {config, MINUTE_IN_MILLIS, HOUR_IN_MILLIS} from '../settings/config';
+import {createShortcuts, removeShortcuts} from './shortcuts';
 
 const logger = getLogger(path.basename(__filename));
 
 const appFolder = path.resolve(process.execPath, '..');
 const rootFolder = path.resolve(appFolder, '..');
 const updateDotExe = path.join(rootFolder, 'Update.exe');
+const exeName = path.basename(process.execPath);
+const exePath = path.join(rootFolder, exeName);
```
atomrc, line 40: This is the main change. This is now how we compute the exec path (starting from the `root` dir of the application, not the current folder of the exec)

PR #6423, opened 2023-02-15
fix: Do not assume base64 encoded string when encrypting/decrypting

atomrc:
With the previous version, we assumed that the decrypted string will be base65 encoded. Which is not necessarily true. We also need to be able to decrypt random bytes

### DIFF
--- a/electron/src/preload/preload-webview.ts
+++ b/electron/src/preload/preload-webview.ts
@@ -243,14 +243,13 @@ process.once('loaded', () => {
     getDesktopSources: opts => ipcRenderer.invoke(EVENT_TYPE.ACTION.GET_DESKTOP_SOURCES, opts),
   };
   global.systemCrypto = {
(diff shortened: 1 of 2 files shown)


electron/src/preload/preload-webview.ts
```diff
@@ -243,14 +243,13 @@ process.once('loaded', () => {
     getDesktopSources: opts => ipcRenderer.invoke(EVENT_TYPE.ACTION.GET_DESKTOP_SOURCES, opts),
   };
   global.systemCrypto = {
-    decrypt: async (encrypted: Uint8Array): Promise<Uint8Array> => {
-      const plainText = await ipcRenderer.invoke(EVENT_TYPE.ACTION.DECRYPT, encrypted);
-      return Decoder.fromBase64(plainText).asBytes;
```
atomrc, line 248: this could cause an issue in case the value that was encrypted is not base64 encoded

PatrykBuniX (approved):

PR #6603, opened 2023-04-17
feat: migrate to ts components

przemvs:
Migrated files from JS to TS.
https://wearezeta.atlassian.net/browse/SQSERVICES-1998

Also a lot of files fixed by eslint & prettier.

PR #6622, opened 2023-04-25
feat: migrate lib js files to ts

przemvs:
Migration of lib JS files to TS.

This will be merged to dev after merge this one to dev: https://github.com/wireapp/wire-desktop/pull/6603 .


electron/renderer/src/lib/__mocks__/localStorage.ts
```diff
@@ -18,6 +18,8 @@
  */
 
 export class LocalStorageMock {
+  store: {[key: string]: string} = {};
+
   constructor() {
     this.store = {};
   }
```
atomrc, line 25: I guess you don't need the constructor now since you initialize the value in the class
```suggestion

```

electron/renderer/src/lib/__mocks__/localStorage.ts
```diff
@@ -18,6 +18,8 @@
  */
 
 export class LocalStorageMock {
+  store: {[key: string]: string} = {};
```
atomrc, line 21: We could use a `Map` for this

electron/renderer/src/lib/__tests__/WindowUrl.spec.ts
```diff
@@ -35,7 +35,6 @@ describe('WindowUrl', () => {
       const customWebApp = 'https://webapp.qa-demo.wire.link?clienttype=permanent';
       try {
         WindowUrl.createWebAppUrl(rendererPage, customWebApp);
-        done.fail();
```
atomrc, line 38: I think this should fail. This is a behavior change

electron/renderer/src/lib/__tests__/localStorage.spec.ts
```diff
@@ -17,17 +17,11 @@
...
 
-// eslint-disable-next-line jest/no-mocks-import
-import {LocalStorageMock} from '../__mocks__/localStorage';
 import {loadState, saveState} from '../localStorage';
 
 describe('localStorage', () => {
-  beforeEach(() => {
-    global.localStorage = new LocalStorageMock();
-  });
-
```
atomrc, line 28: do we have local storage now? 
In this case, can we remove the mock altogether?

tlebon (approved):
seems fine to me

PR #7047, opened 2023-09-18
feat: Upgrade to electron 26

atomrc:
This brings back the native macos bar as, since version 23 of electron, the `app-region: drag` has changed behavior and we cannot have a drag region on top of some action button inside of a `webview`.  (see https://www.electronjs.org/docs/latest/breaking-changes#behavior-changed-draggable-regions-on-macos)

The idea is to get the electron update ASAP, and eventually design a better drag region embed inside the app later on


PatrykBuniX (approved):
Looks good 😎

PR #7057, opened 2023-09-19
fix: remove logs for deleted accounts, remove unused endpoints [WPB-1523]

tlebon:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

Deletes current logs when a user removes an account to remove history of the deleted user from the computer.

### DIFF
--- a/electron/src/lib/LocalAccountDeletion.ts
+++ b/electron/src/lib/LocalAccountDeletion.ts
@@ -19,13 +19,13 @@
 
 import {app, Session, webContents} from 'electron';
 import * as fs from 'fs-extra';
+import {truncate} from 'lodash';
 
 import * as path from 'path';
 
 import {ValidationUtil} from '@wireapp/commons';
 
 import {getLogger} from '../logging/getLogger';
-import {getLogFilenames} from '../logging/loggerUtils';
 
 const USER_DATA_DIR = app.getPath('userData');
 const LOG_DIR = path.join(USER_DATA_DIR, 'logs');
@@ -39,6 +39,8 @@ const clearStorage = async (session: Session): Promise<void> => {
 };
 
 export async function deleteAccount(id: number, accountId: string, partitionId?: string): Promise<void> {
+  const truncatedId = truncate(accountId, {length: 5});
+
   // Delete session data
   try {
     const webviewWebContent = webContents.fromId(id);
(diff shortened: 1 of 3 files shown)


atomrc (approved):

electron/src/lib/LocalAccountDeletion.ts
```diff
@@ -77,8 +78,13 @@ export async function deleteAccount(id: number, accountId: string, partitionId?:
...
+    const relativeFilePaths = getLogFilenames();
     await fs.remove(sessionFolder);
-    logger.log(`Deleted logs folder for account "${accountId}".`);
+    relativeFilePaths.forEach(async log => {
+      const logPath = path.join(LOG_DIR, log);
+      await fs.remove(logPath);
+    });
+    logger.log(`Deleted logs folder for current account.`);
   } catch (error: any) {
     logger.error(`Failed to delete logs folder for account "${accountId}", reason: "${error.message}".`);
```
atomrc, line 89: Shouldn't we change that as well?

tlebon: In this case I guess logs probably weren't deleted so I thought good (or acceptable) to leave it here? Idk I'm open either way

tlebon: I'd be open to leave an extended warning here to tell the user to manually delete their logs, with the location. But idk if that's best

atomrc: Thing is, this log will go to the main `electron.log` isn't that correct. I guess we should never put any account ID in those logs (that's my understanding)

tlebon: So I can imagine that in some cases it makes sense to have userId logs in this file. But just to clear them when the user is deleted. But the ticket does indeed say "don't log userId"

atomrc: wait, are you removing the entire `electron.log` and `electron.old` in this PR? Might be a little overkill 🤔 
We are going to lose logs for other accounts

atomrc (approved):

PR #7106, opened 2023-10-04
fix: Improve how webapp environments are displayed in the dev menu

atomrc:
This is the environments menu we currently have 

image

This PR will simplify this menu and only allow 3 envs:
- production
- beta (currently staging, the one used by Internal wrapper by default)
- edge (bleeding edge code straight from dev, but the courageous ones)

image


V-Gira (approved):

PatrykBuniX (approved):
Looks good 😎

PR #7793, opened 2024-04-23
feat: picture in picture calling view PoC

PatrykBuniX:
This PR adds support for Picture-In-Picture calling view PoC.

### DIFF
--- a/electron/src/calling/PictureInPictureCall.ts
+++ b/electron/src/calling/PictureInPictureCall.ts
@@ -27,9 +27,8 @@ export const isPictureInPictureCallWindow = (frameName: string): boolean => {
 
 export const getPictureInPictureCallWindowOptions = (): Electron.BrowserWindowConstructorOptions => {
   return getNewWindowOptions({
-    title: 'Calling UI',
-    width: 290,
-    height: 290,
+    width: 1026,
+    height: 829,
     resizable: true,
     fullscreenable: true,
     maximizable: true,
--- a/electron/src/calling/PictureInPictureCall.ts
+++ b/electron/src/calling/PictureInPictureCall.ts
@@ -17,8 +17,21 @@
  *
  */
 
+import {getNewWindowOptions} from '../window/WindowUtil';
+
 const PICTURE_IN_PICTURE_CALL_FRAME_NAME = 'WIRE_PICTURE_IN_PICTURE_CALL';
 
 export const isPictureInPictureCallWindow = (frameName: string): boolean => {
   return frameName === PICTURE_IN_PICTURE_CALL_FRAME_NAME;
+};
+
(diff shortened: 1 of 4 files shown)


electron/src/calling/PictureInPictureCall.ts
```diff
@@ -0,0 +1,36 @@
...
+const PICTURE_IN_PICTURE_CALL_FRAME_NAME = 'WIRE_PICTURE_IN_PICTURE_CALL';
+
+export const isPictureInPictureCallWindow = (frameName: string): boolean => {
+  return frameName === PICTURE_IN_PICTURE_CALL_FRAME_NAME;
+};
+
+export const getPictureInPictureCallWindowOptions = (): Electron.BrowserWindowConstructorOptions => {
+  return getNewWindowOptions({
+    width: 1026,
+    height: 829,
```
aweiss-dev, line 31: Not a fan of the fixed values here. Is there a way to get these values from the open window?

electron/src/window/WindowUtil.ts
```diff
@@ -78,3 +78,62 @@ export const openExternal = async (url: string, httpsOnly: boolean = false): Pro
     logger.error(error);
   }
 };
+
```
aweiss-dev, line 81: I like this a lot 🤍

electron/src/calling/PictureInPictureCall.ts
PatrykBuniX: Not sure what do you mean? Here we specify, what is the initial size of the newly opened window. User can then resize it to match their needs.

aweiss-dev: I mean, hat we open it in an fixed size. I would rather see us inherit this size from the parent window (like calc(inherit-50px) ). Something like that :D

PatrykBuniX: This is a product decision I'd say. This is just the first iteration, it's likely that we will change things here and there

aweiss-dev: Fine for me

aweiss-dev (approved):

electron/src/calling/PictureInPictureCall.ts
tlebon: i tend to agree with adrian somewhat. can we allow the size to go smaller than this? i feel like theres a scenario id want it to be like a small box i could just float somewhere innocuous.

tlebon (approved):

electron/src/calling/PictureInPictureCall.ts
PatrykBuniX: Yes, this will be improved in the future, the current designs for MVP v2 don't address smaller sizes.

PR #7843, opened 2024-05-14
Revert "build(deps): bump electron-dl from 3.5.2 to 4.0.0"

tlebon:
Reverts wireapp/wire-desktop#7816


tlebon:
this is blocked because it is using an esm module. see wpb-9121.

PR #7935, opened 2024-06-11
runfix: don't clear cache by default when closing the app

PatrykBuniX:
# What's new in this PR?

We don't want to clear cache every time the app is closed. Because of that we had to download all the javascript every time the app was lunched.

### DIFF
--- a/electron/src/runtime/lifecycle.ts
+++ b/electron/src/runtime/lifecycle.ts
@@ -54,7 +54,7 @@ export const checkSingleInstance = async () => {
     logger.info('Checking if we are the first instance ...', isFirstInstance);
 
     if (!EnvironmentUtil.platform.IS_WINDOWS && !isFirstInstance) {
-      await quit(false);
+      await quit();
     } else {
       app.on('second-instance', () => WindowManager.showPrimaryWindow());
     }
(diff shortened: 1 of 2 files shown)


electron/src/update/squirrel.ts
```diff
@@ -122,24 +122,26 @@ async function scheduleUpdate(): Promise<void> {
...
       return;
     }
 
     case SQUIRREL_EVENT.OBSOLETE: {
-      await lifecycle.quit();
+      // This is called when the app is updated but the old version is still running
+      await lifecycle.quit(true);
       return;
     }
   }
```
PatrykBuniX, line 147: It's still fine to clear the cache if new version of app is installed (we will need to download it anyway)

electron/src/runtime/lifecycle.ts
```diff
@@ -83,7 +83,7 @@ export const addRelaunchListeners = (listener: () => void) => {
   relaunchListeners.push(listener);
 };
 
-export const quit = async (clearCache = true): Promise<void> => {
+export const quit = async (clearCache = false): Promise<void> => {
```
atomrc, line 86: So I'm guessing in most cases we were actually clearing the cache because there was not arguments, right?

PatrykBuniX: Correct, there was only one case where we'd not clear it

atomrc (approved):

PR #8242, opened 2024-10-02
feat: hide menu on calling pop out window [WPB-10680]

e-maad:
## Description

Hide menu on pop out window.


V-Gira (approved):
confirmed to work locally :+1:

PR #8392, opened 2024-12-04
chore: fix the annoying dock bounce on macOS (WPB-14749)

emil-wire:
# What's new in this PR?

### Issues

After a recent electron update, the dock icon will bounce relentlessly until the focus is on the window again. Previously it bounced once

### Causes (Optional)

https://github.com/electron/electron/pull/41391
The change in electron was made to align behaviour across platforms. 

### Solutions

Added a platform check to preserve the current behaviour on windows, but change the behaviour on macos to bouncing once.

### DIFF
--- a/electron/src/menu/TrayHandler.ts
+++ b/electron/src/menu/TrayHandler.ts
@@ -96,7 +96,7 @@ export class TrayHandler {
     this.trayIcon?.setToolTip(config.name);
   }
 
- private flashApplicationWindow(win: BrowserWindow, count?: number): void {
+  private flashApplicationWindow(win: BrowserWindow, count?: number): void {
     if (win.isFocused() || !count) {
       if (process.platform === 'win32') {
         win.flashFrame(false);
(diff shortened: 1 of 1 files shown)


aweiss-dev (approved):
Good job :)

electron/src/menu/TrayHandler.ts
```diff
@@ -98,8 +98,19 @@ export class TrayHandler {
...
+      return;
+    }
+    /* After an Electron API change https://github.com/electron/electron/pull/41391
+       flashFrame() leads to a constant bouncing of the dock icon on macOS.
+       By calling the dock.bounce() directly, we avoid this behavior. the "informational"
+       is optional (default), but makes it easier to read
+    */
+    if (process.platform === 'darwin') {
+      app.dock.bounce('informational');
+    } else if (process.platform === 'win32') {
```
V-Gira, line 113: woops we missed this

```suggestion
    } else if (count > this.lastUnreadCount) {
```

thisisamir98 (approved):

PR #8733, opened 2025-04-29
fix(macOS): fix entitlements to remove server

aweiss-dev:
The server entitlement seems to be unnecessary and Apple flagged it. To unblock a new release we remove it in this PR.

### DIFF
--- a/app-config/package.json
+++ b/app-config/package.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "wire-web-config-internal": "https://github.com/wireapp/wire-web-config-default#v0.34.1",
-    "wire-web-config-production": "https://github.com/wireapp/wire-web-config-wire#v0.34.1"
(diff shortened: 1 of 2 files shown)


app-config/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "wire-web-config-internal": "https://github.com/wireapp/wire-web-config-default#v0.34.1",
-    "wire-web-config-production": "https://github.com/wireapp/wire-web-config-wire#v0.34.1"
+    "wire-web-config-internal": "https://github.com/wireapp/wire-web-config-default#v0.34.3-0",
```
aweiss-dev, line 3:
```suggestion
    "wire-web-config-internal": "https://github.com/wireapp/wire-web-config-default#v0.34.2",
```

V-Gira: was about to say

V-Gira (approved):
Please target `main` if it's intended for a hotfix build

PR #9114, opened 2025-09-29
fix: sandboxed module resolution [WPB-20012]

vkuprin:
Fixes context isolation implementation to work properly in sandboxed preload environment.

• **Fix sandboxed module resolution**: Electron's sandboxed preload scripts have very limited CommonJS module resolution capabilities
• **Replace shared imports with inline constants**: Sandboxed context with contextIsolation: true and nodeIntegration: false cannot import relative modules
• **Eliminate duplicate constants**: Consolidate EVENT_TYPE constants that were duplicated across multiple files due to import restrictions
• **Add security documentation**: Comprehensive comments explaining why inline constants are required for context isolation
• **Fix linting violations**: Proper JSDoc documentation and ESLint compliance for all inline implementations

### DIFF
--- a/electron/src/main.ts
+++ b/electron/src/main.ts
@@ -160,7 +160,8 @@ let isFullScreen = false;
 let isQuitting = false;
 let main: BrowserWindow;
 
-(async () => { // NOSONAR
+// eslint-disable-next-line prettier/prettier
+(async () => { // NOSONAR - TypeScript config doesn't support top-level await, IIFE pattern required
   for (const [key, value] of Object.entries(config)) {
     if (value === undefined || (typeof value === 'number' && Number.isNaN(value))) {
       logger.warn(`Configuration key "${key}" not defined.`);


electron/src/preload/preload-app.ts
```diff
@@ -18,19 +18,200 @@
...
+// These are kept in sync with ../shared/contextIsolationConstants.ts
 
-import {WebAppEvents} from '@wireapp/webapp-events';
+/**
+ * Event type constants for IPC communication
+ *
+ * IMPORTANT: This is a necessary duplication of ../shared/contextIsolationConstants.ts
+ * due to Electron sandbox limitations. Keep in sync with the shared file.
+ */
+const EVENT_TYPE = {
```
aweiss-dev, line 33: no one will remember to keep this in sync with the package. 
it also defeats the use case of the package. 

This needs to be done automated, if the contents need to be copied over from the package, then there should be a sync command that does it automatically aka "yarn sync:events"

bin/sync-events.js
```diff
@@ -0,0 +1,261 @@
...
+
+/**
+ * Generate the shared constants file content
+ * @param {string} eventTypeObject - The EVENT_TYPE object string
+ * @returns {string} Complete file content
+ */
+function generateSharedConstantsContent(eventTypeObject) {
+  return `/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
aweiss-dev, line 86: propably should be 2025, since the file is new

aweiss-dev (approved):

PR #9358, opened 2025-12-15
feat: Electron enable context isolation

adamlow-wire:
# Pull Request

## Summary

This PR enables **Electron context isolation** for the main window.

### What Changed

1. **Enabled context isolation** (`contextIsolation: true`) in main window webPreferences
2. **Refactored preload scripts** to use `contextBridge.exposeInMainWorld()` instead of direct window property assignments
3. **Updated all renderer code** to use `window.electronAPI.*` instead of direct window properties
4. **Fixed context menu actions** to work correctly with webviews by executing actions directly on webContents
5. **Added error handling** for account removal when webviews aren't fully initialized
6. **Fixed locale timing issue** by replacing broken setTimeout approach with a getter function

### Testing

Manual testing of context menu's, account addition/removal, can't test SSO as don't have the setup to do so.
---

## Security Checklist (required)

- [x] **External inputs are validated & sanitized** on client and/or server where applicable.
  - IPC handlers validate inputs before processing
  - Account IDs are validated before operations
- [x] **API responses are validated**; unexpected shapes are handled safely (fallbacks or errors).
  - Error handling added for account removal edge cases
  - Try-catch blocks around webContents operations
- [x] **No unsafe HTML is rendered**; if unavoidable, sanitization is applied **and** documented where it happens.
  - No HTML rendering changes in this PR
- [x] **Injection risks (XSS/SQL/command) are prevented** via safe APIs and/or escaping.
  - Context isolation itself prevents XSS by isolating contexts
  - IPC communication uses typed channels, not string concatenation

## Standards Acknowledgement (required)

- [x] I have read and this PR **upholds** our [Coding Standards](https://github.com/wireapp/wire-desktop/tree/docs/coding-standards.md) and [Tech Radar Choices](https://github.com/wireapp/wire-desktop/tree/docs/tech-radar.md).

---

## Screenshots or demo (if the user interface changed)

No UI changes - this is a security hardening change that maintains existing functionality.

---

### Testing Checklist

- [x] Context menus work (text, image, link, selection)
- [x] Account operations work (add, remove, logout)
- [ ] SSO flow works
- [x] Locale/language switching works
- [x] No console errors
- [x] IPC communication works correctly


EnricoSchw (commented):
Hi @adamlow-wire , 

thx for this PR, we need to test this change thoroughly. Because If we place the app in an isolated context, this change will affect our calling.

Please note that audio and video streams are singletons. The isolated context may make them inaccessible in parts of our app. Therefore, we need to test this changes.

Furthermore, our app runs in a sandbox on Mac, which maybe makes things more complicated.

That's why I built the app locally, signed it, and launched it in a sandbox. Unfortunately, I couldn't get past the login screen. Here the error I see. 

Bildschirmfoto 2025-12-18 um 09 28 00

Unfortunately, it's not very informative, but maybe you can make something of it.

PR #9451, opened 2026-01-30
feat(mdm): initial MDM support for Windows and macOS [WPB-23215]

adamlow-wire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23215" title="WPB-23215" target="_blank">WPB-23215</a>  [Desktop] - Add MDM support to Wire Desktop for MacOS & Windows
  </td></table>
  <br />
 

# Pull Request

## Summary

- **What did I change and why?**  
  Enterprise managed configuration (MDM) for Wire Desktop: admins can enforce settings via Windows Registry (Group Policy or user key) or macOS managed preferences. Managed values override user settings and are cached for the process lifetime. Config is read in the main process only; the renderer receives a redacted config via IPC (no registry/OS access in the webview). Implemented: safe download path with home-dir traversal check on Windows and macOS; locale normalization to supported language keys; registry key sanitization (alphanumeric app name) for Windows paths; proxy credentials redacted before exposing to renderer. Removed `environment` from MDM (use `webappUrl` plus build type instead). Added unit tests for parsing, validation, redaction, and a test-only override seam. Added `docs/mdm-managed-configuration.md` for MDM/Group Policy admins and testing notes.

- **Risks and how to roll out / roll back (e.g. feature flags):**  
  Low risk: MDM is opt-in by deployment (admins deploy registry/preferences). No feature flag. If admins remove policy keys or managed preferences, users restart the app and settings fall back to user/default. Roll back: remove deployed registry keys or managed preferences and restart clients.

---

## Security Checklist (required)

- [x] **External inputs are validated & sanitized** on client and/or server where applicable.  
  Registry/preference values validated (Joi schema); app name sanitized for Windows registry path (alphanumeric); download path validated and checked for traversal outside user home.
- [x] **API responses are validated**; unexpected shapes are handled safely (fallbacks or errors).  
  N/A for this PR (config source is OS, not an API). Managed config shape validated via schema; invalid values dropped with log.
- [x] **No unsafe HTML is rendered**; if unavoidable, sanitization is applied **and** documented where it happens.  
  No HTML from managed config; no UI change that renders user/OS input as HTML.
- [x] **Injection risks (XSS/SQL/command) are prevented** via safe APIs and/or escaping.  
  Registry key path built from sanitized app name only; download path resolved and checked for traversal; proxy URL not executed as command.

## Standards Acknowledgement (required)

- [x] I have read and this PR **upholds** our [Coding Standards](https://github.com/wireapp/wire-desktop/tree/docs/coding-standards.md) and [Tech Radar Choices](https://github.com/wireapp/wire-desktop/tree/docs/tech-radar.md).

---

## Screenshots or demo (if the user interface changed)

No UI changes. Managed settings affect behaviour (e.g. download location, locale, menu bar visibility) but do not add new screens.

## Notes for reviewers

- **Trade-offs:** Single-load cache (no runtime refresh of managed config); config changes require app restart. Kept current consumer pattern (import `ManagedConfig` / IPC) rather than a full “single owner” refactor to limit scope.
- **Follow-ups (linked issues):** None.
- **Linked PRs (e.g. web-packages):** None.


electron/src/main.ts
```diff
@@ -99,31 +100,64 @@ const customProtocolHandler = new CustomProtocolHandler();
...
+    return null;
+  }
+  const home = app.getPath('home');
+  const resolved = path.resolve(home, trimmed);
+  const normalizedHome = path.resolve(home);
+  const normalizedResolved = path.resolve(resolved);
+  if (!normalizedResolved.startsWith(normalizedHome + path.sep) && normalizedResolved !== normalizedHome) {
+    return null;
+  }
+  return normalizedResolved;
```
screendriver, line 129: I understand (at least I believe) that the user is not able to download in a directory like `../../../../foo` when it escapes your home directory. But the implementation is very hard to read and understand. Could this be refactored to make it more readable? Also an isolated test just for this (very important?) functionality would be great 👍

electron/src/main.ts
```diff
@@ -99,31 +100,64 @@ const customProtocolHandler = new CustomProtocolHandler();
...
-        }),
-        buttons: [locale.getText('enforcedDownloadButton')],
-      });
-    },
-  });
+/** Managed download path is supported on Windows and macOS only (user-content download folder). */
+const useManagedDownloadPath =
+  (EnvironmentUtil.platform.IS_WINDOWS || EnvironmentUtil.platform.IS_MAC_OS) &&
+  typeof customDownloadPath === 'string' &&
+  customDownloadPath.length > 0;
```
screendriver, line 138: I think we have a bug in here: before `customDownloadPath` was also set when running on Linux and the condition evaluated to `true`. But now it evaluates to `false` when running on Linux because it only checks for Windows and macOS.

If Linux builds previously relied on `init.json` `downloadPath` being enforced that no longer happens now.

electron/src/main.ts
```diff
@@ -99,31 +100,64 @@ const customProtocolHandler = new CustomProtocolHandler();
...
+
+if (useManagedDownloadPath) {
+  const safeDir = getSafeDownloadDirectory(customDownloadPath);
+  if (safeDir) {
```
screendriver, line 142: This would mean when you use `Linux` and don't have a managed device, you are not allowed to download the application in different directory outside of `/home`, right? (this was possible before)

electron/src/menu/system.ts
```diff
@@ -174,7 +178,11 @@ const toggleFullScreenTemplate: MenuItemConstructorOptions = {
 
 const toggleAutoLaunchTemplate: MenuItemConstructorOptions = {
   checked: settings.restore(SettingsType.AUTO_LAUNCH, false),
+  enabled: !isSettingManaged(SettingsType.AUTO_LAUNCH),
   click: () => {
+    if (isSettingManaged(SettingsType.AUTO_LAUNCH)) {
+      return;
+    }
```
screendriver: 2 lines above we are already checking if the menu entry should be enabled or not. If it is disabled the user is not able to click anything. So the check is done twice I guess.

electron/src/preload/preload-webview.ts
```diff
@@ -257,7 +257,7 @@ function reportWebappAVSVersion(): void {
 const _clearImmediate = clearImmediate;
 const _setImmediate = setImmediate;
 
-process.once('loaded', () => {
+process.once('loaded', async () => {
```
screendriver, line 260: [process.once('loaded')](https://nodejs.org/api/process.html) does not allow to put in functions that returns [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). All of the code in there must be synchronous. The reason why nothing complains here is because of the lax compiler / linter settings in this repository.

It would also have the side effect that `global.desktopAppConfig` is set after an async round‑trip (it was synchronous before). If the application expects `desktopAppConfig` to be available immediately at document start this could break on Linux without MDM.

electron/src/menu/system.ts
```diff
@@ -488,11 +504,40 @@ export const toggleMenuBar = (): void => {
...
+  try {
+    if (shouldAutoLaunch) {
+      await launcher.enable();
+    } else {
+      await launcher.disable();
+    }
+  } catch (error: unknown) {
+    const message = error instanceof Error ? error.message : String(error);
+    const isLoginItemNotFound =
+      message.includes('-1728') || message.includes("Can't get login item") || message.includes('cannot be read');
```
screendriver, line 533: What is the meaning of `-1728`? I think in this context it is a macOS error code. I am right? It would be better to extract this value into a meaningful self describing variable otherwise no one in the future will know what's the meaning of it.

PR #9660, opened 2026-06-17
Extract login into re-usable action [WPB-26068]

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26068" title="WPB-26068" target="_blank">StoryWPB-26068</a>  [Desktop/QA] Create critical flow test for calling
  </td></table>
  <br />
 

Refactoring for the login function as this util will be needed for many other tests


e2e-tests/actions/loginUser.ts
screendriver, line 1: Small naming suggestion: could we avoid putting this under `utils`?

I usually try to avoid names like `utils` or `helpers`, because they tend to become generic dumping grounds over time. This function is more specific than that: it represents a login flow used by the E2E tests.

What do you think about moving it to something like:

> e2e-tests/flows/login.ts

or, if we want to be even more explicit:

> e2e-tests/flows/logInUser.ts

markbrockhoff: Good point, but I think I'd prefer something like "actions" over "flows" as flows kind of sounds like full flows from a users perspective while the goal is to have multiple composable functions / actions which group interactions on POMs. I fear that "flows" could cause the contents of the folder to become too "smart" :D

screendriver: Also fine 🙂 I don’t have a strong opinion here (except `util` 🤪)

screendriver (approved):

PR #9675, opened 2026-06-24
feat: pull MDM config and check if managed [WPB-26373]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26373" title="WPB-26373" target="_blank">SpikeWPB-26373</a>  [Web] POC for differentiating Managed and Own devices on the web
  </td></table>
  <br />
 

### Description

Pulled MDM config to pass `isManaged' to webApp.


thisisamir98 (commented):
Looked through this — nice POC overall and the README for local testing is helpful. I left a couple of comments on things that might need a product/MDM check before merge.

electron/src/managed/backends/windows.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return subkeysOf(registry, HKEY.HKEY_LOCAL_MACHINE, WINDOWS_CLOUD_DOMAIN_JOIN_KEY).length > 0;
+}
+
+export function isDeviceManagedWindows(): boolean {
+  const registry = loadRegistry();
+  if (!registry) {
+    return false;
+  }
```
thisisamir98, line 109: On Windows we treat MDM enrollment or Azure AD join the same as having `applockOverride` set in `Policies\Wire`. That means a normal company laptop with Intune but no Wire policy will still send `applockOverride: true` to the webapp and force app-lock on.

Is that what product wants for this spike? The ticket sounds more like telling managed vs own device apart — if that is the goal, maybe enrollment should be a separate signal and not always turn on the override.

package.json
```diff
@@ -30,6 +30,7 @@
     "redux": "4.2.1",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.4.2",
+    "registry-js": "1.16.1",
```
thisisamir98: `registry-js` is Windows-only native code. We lazy-require it so runtime on mac/linux is fine, but putting it in regular `dependencies` can still make `yarn install` fail on non-Windows CI or dev machines when prebuilds do not match.

Worth moving to `optionalDependencies` (same pattern as `fsevents`) so mac/linux builds do not depend on it compiling.

electron/src/managed/ManagedConfig.test.main.ts
```diff
@@ -0,0 +1,70 @@
+/*
```
zskhan, line 1: Should we add tests for macOS and Windows as well, or is the Linux test sufficient?

electron/src/managed/backends/windows.ts
e-maad: Yes, you are right. The main goal was to identify if the device is managed or not but there is a catch. We can't check that on electron (without shell commands). So, we decided to have an MDM config to control the behavior of the app. In future, if we want to push other configs we can use the same path.

electron/src/managed/ManagedConfig.test.main.ts
e-maad: Nice catch, I missed it. Adding it now.

thisisamir98 (approved):

PR #9676, opened 2026-06-25
test: add e2e regression tests for multiple accounts [WPB-26365]

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26365" title="WPB-26365" target="_blank">StoryWPB-26365</a>  [Desktop/QA] Write Multiple Accounts regression tests
  </td></table>
  <br />
 

This PR adds a new test suite for multiple accounts.

### DIFF
--- a/e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
+++ b/e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
@@ -19,8 +19,11 @@
 
 import {createGroup} from '../../actions/createGroup';
 import {loginUser} from '../../actions/loginUser';
+import {sendConnectionRequest} from '../../actions/sendConnectionRequest';
 import {expect, test} from '../../fixtures';
 import {accountsSidebar} from '../../poms/app/accountsSidebar.page';
+import {connectionRequestPage} from '../../poms/webapp/connectionRequest.page';
+import {conversation} from '../../poms/webapp/conversation.page';
 import {conversationsList} from '../../poms/webapp/conversationList.page';
 import {conversationsSidebar} from '../../poms/webapp/conversationsSidebar.page';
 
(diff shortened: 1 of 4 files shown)


screendriver (approved):

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+      const [userA, userAPage] = [userATeam.owner, app.page];
+      const userAGroup = 'User A Group';
+
+      const userBTeam = await createTeam('Multiple Accounts B', {users: []});
+      const userB = userBTeam.owner;
+      const userBGroup = 'User B Group';
+
+      await test.step('UserA logs in', async () => {
+        await loginUser(userAPage, userA);
+        await expect(conversationsSidebar(userAPage).userAvatar).toContainText(userA.initials);
```
markbrockhoff, line 47: You don't need to wait for the login to finish, since the login step is a very common one it will wait for the login to finish before continuing using the LOGIN_TIMEOUT internally.

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+test.describe('Multiple Accounts', async () => {
+  test(
+    'I want to switch between multiple accounts and see correct conversation',
+    {tag: ['@TC-11002', '@regression']},
+    async ({app, createTeam}) => {
+      const userATeam = await createTeam('Multiple Accounts A', {users: []});
+      const [userA, userAPage] = [userATeam.owner, app.page];
+      const userAGroup = 'User A Group';
+
+      const userBTeam = await createTeam('Multiple Accounts B', {users: []});
```
markbrockhoff, line 41: Why are you using two teams? Wouldn't it be possible to just use two users or is there a reason for it?

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+import {LOGIN_TIMEOUT} from '../../poms/webapp/login.page';
+import {ssoPage} from '../../poms/webapp/sso.page';
+
+test.describe('Multiple Accounts', async () => {
+  test(
+    'I want to switch between multiple accounts and see correct conversation',
+    {tag: ['@TC-11002', '@regression']},
+    async ({app, createTeam}) => {
+      const userATeam = await createTeam('Multiple Accounts A', {users: []});
+      const [userA, userAPage] = [userATeam.owner, app.page];
```
markbrockhoff, line 38: Please don't extract `app.page` here. The test only uses one page which is the one of the app. If you copy the reference to the first page here assertions on it will always be true, no matter if the switch between accounts worked or not.

```suggestion
      const userA = userATeam.owner;
```

Just use `app.page` instead of `userAPage`

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+
+      await test.step('UserB creates empty group', async () => {
+        await createGroup(app.page, userBGroup, []);
+      });
+
+      await test.step('UserB switches account back to A', async () => {
+        await accountsSidebar(app).switchAccount(0);
+      });
+
+      await test.step('UserA does not see group conversation of UserA', async () => {
```
markbrockhoff, line 72:
```suggestion
      await test.step('UserA does not see group conversation of UserB', async () => {
```

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+    await test.step('UserA switches back to main account', async () => {
+      await accountsSidebar(app).switchAccount(0);
+    });
+
+    await test.step('UserA removes active main account', async () => {
+      await accountsSidebar(app).removeAccount(0);
+      await expect(accountsSidebar(app).getAccount(userB)).toBeVisible();
+      await expect(accountsSidebar(app).getAccount(userA)).not.toBeVisible();
+      await expect(accountsSidebar(app).accountItems).toHaveCount(1);
+    });
```
markbrockhoff, line 148: Maybe it would be a good idea to also assert on something within the accounts webview to verify the switch after deletion actually worked. Otherwise the account might be gone from the sidebar but still be shown:


```suggestion
    await test.step('UserA removes active main account', async () => {
      await accountsSidebar(app).removeAccount(0);
      await expect(accountsSidebar(app).getAccount(userB)).toBeVisible();
      await expect(accountsSidebar(app).getAccount(userA)).not.toBeVisible();
      await expect(accountsSidebar(app).accountItems).toHaveCount(1);
      await expect(conversationsSidebar(app.page).userAvatar).toContainText(userB.initials);
    });
```

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: So you suggest to get rid of the assertion afterwards, did I get that right? 🤔

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: Because the crit flow test for multiple accounts tests similar behavior but with regular users, I thought it would be good to have here team users. But we totally could use regular users, what do you think? :)

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: Good catch, thanks!

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: ✅

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: ✅

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
markbrockhoff: Exactly, it simply doesn't add value to this test case :)

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
markbrockhoff: Hmm, well that actually raises the question of how useful this test even is. Now that I look at the critical flow one I do have to say they are very similar...
To be honest I'm even questioning whether we need this test case inside the suite. What do you think, could it make sense to simply delete it to not duplicate the critical flow one?

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: ✅

e2e-tests/specs/multipleAccounts/multipleAccountsSuite.spec.ts
JacquelineLehner: Yes, agreed. I will remove it since it's too similar 🗑️

PR #9681, opened 2026-06-30
Refresh About dialog versions without restarting desktop [WPB-26554]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26554" title="WPB-26554" target="_blank">StoryWPB-26554</a>  [Desktop/QA] Write Notification regression tests
  </td></table>
  <br />
 

## What changed

This updates the About dialog so that it requests the currently loaded webapp and AVS versions from the active webview before rendering the dialog.

The previously known values are still kept as a fallback, so the dialog can continue to open even when no webview is available, the webview is not ready, or the version request times out.

## Why

The About dialog could show stale version information after the webapp was updated or reloaded inside the desktop wrapper. The values only became correct after restarting the whole Electron process.

That is confusing for internal builds, because `WireInternal -> About WireInternal` should describe the currently running webapp, not the version that was active when the desktop process started.

## Notes

This only affects the displayed webapp and AVS versions in the About dialog.
It does not change the desktop app version or the update flow itself.

### DIFF
--- a/electron/src/preload/menu/preload-about.test.renderer.ts
+++ b/electron/src/preload/menu/preload-about.test.renderer.ts
@@ -44,7 +44,7 @@ describe('loadedAboutScreen', () => {
       webappAVSVersion: '9.0.test',
     });
 
-    assert.ok(sendSpy.calledOnceWith(EVENT_TYPE.ABOUT.LOCALE_VALUES, []));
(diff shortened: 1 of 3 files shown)


e-maad (approved):
LGTM 🚀

PR #9685, opened 2026-07-01
test: write Menu Bar regression tests [WPB-26320]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26320" title="WPB-26320" target="_blank">StoryWPB-26320</a>  [Desktop/QA] Write Menu Bar regression tests
  </td></table>
  <br />
 

This PR introduces automated test suites for testing the Electron Application Menu bar, its connected actions, and keyboard shortcut accelerators.

### DIFF
--- a/e2e-tests/specs/regression/menuBar.spec.ts
+++ b/e2e-tests/specs/regression/menuBar.spec.ts
@@ -128,6 +128,36 @@ test.describe('Menu Bar', () => {
   });
 
   test(
+    'Delete conversation content with menu bar',
+    {tag: ['@TC-11042', '@regression']},
+    async ({app, createUser, createTeam, createPage}) => {
+      const userB = await createUser();
+      const {owner: userA} = await createTeam('Test Team', {users: [userB]});
+      const userAPage = app.page;
+      const userBPage = await createPage();
+
+      await Promise.all([loginUser(userAPage, userA), loginUser(userBPage, userB)]);
+      await connectWithUser(userAPage, userB);
+
+      await test.step('User A can delete 1:1 conversation', async () => {
+        await conversationsList(userAPage).getConversation(userB.fullName).open();
+        await triggerApplicationMenu(app, ['Delete Content...']);
+
+        await expect(app.page.getByRole('dialog').getByText('Clear content?')).toBeVisible();
+        await app.page.getByRole('dialog').getByRole('button', {name: 'Cancel'}).click();
+      });
+
+      await test.step('User A can ping group conversation', async () => {
+        await createGroup(app.page, 'Test group', [userB]);
+        await conversationsList(userAPage).getConversation('Test group').open();
+
+        await triggerApplicationMenu(app, ['Delete Content...']);
+        await expect(app.page.getByRole('dialog').getByText('Clear content?')).toBeVisible();
+      });
+    },
+  );
+
+  test(
     'Verify I can ping 1:1 and group conversation using the menu bar',
     {tag: ['@TC-11043', '@regression']},
     async ({app, createUser, createTeam, createPage}) => {
--- a/e2e-tests/specs/regression/menuBar.spec.ts
+++ b/e2e-tests/specs/regression/menuBar.spec.ts
@@ -17,30 +17,30 @@
  *
  */
 
+import {MenuItem} from 'electron';
+
+import {conversationsList} from './../../poms/webapp/conversationList.page';
+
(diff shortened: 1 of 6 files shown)


e2e-tests/specs/regression/menuBar.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      },
+      verifyGroup: async (page: Page) => {
+        await expect(callCell(page)).toBeVisible();
+      },
+    },
+    {
+      name: 'Verify opening people popover with menu bar in the conversation',
+      tag: '@TC-11045',
+      menuItem: 'People',
+      expectedAccelerator: 'CmdOrCtrl+I',
```
markbrockhoff: 🎨  Just a thought, but all the other test cases of this parameterized test have effectively the same verification for both direct and group. This one is the only one which requires special treatment. Also considering the extra prop to verify the accelerator here.
Maybe it would be cleaner to not have it as part of the parameterized test but just be its own specialized test case? What do you think?

e2e-tests/specs/regression/menuBar.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      const userBPage = await createPage();
+
+      await Promise.all([loginUser(userAPage, userA), loginUser(userBPage, userB)]);
+      await connectWithUser(userAPage, userB);
+
+      await createGroup(app.page, 'Test group', []);
+      await expect(conversationsList(app.page).items).toHaveCount(2);
+
+      await test.step('Archive in 1:1 conversation', async () => {
+        await conversationsList(app.page).getConversation(userB.fullName).open();
```
markbrockhoff: 🔧 To avoid flake it might be a good idea to specify the protocol of the conversation when opening it. Otherwise we might run into the issue where the conversation is removed and re-added during the upgrade ^^'

```suggestion
        await conversationsList(app.page).getConversation(userB.fullName, {protocol: 'mls').open();
```

e2e-tests/specs/regression/menuBar.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      const user = await createUser();
+      await loginUser(app.page, user);
+
+      await createGroup(app.page, 'Test group', []);
+      await conversationsList(app.page).getConversation('Test group').open();
+
+      const menuItem = await triggerApplicationMenu(app, ['Add People...']);
+
+      expect(menuItem.accelerator).toBe('Shift+CmdOrCtrl+K');
+      await expect(app.page.locator('#add-participants')).toBeVisible();
```
markbrockhoff: 🔧  I think we could use an accessible locator here instead of relying on some elements id:

```suggestion
      await expect(app.page.getByRole('complementary').filter({hasText: 'Add participants'})).toBeVisible();
```

e2e-tests/specs/regression/menuBar.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+  return await app.evaluate(async ({Menu, BrowserWindow}, targets) => {
+    const menu = Menu.getApplicationMenu();
+
+    const target = menu?.items.flatMap(item => item.submenu?.items ?? []).find(item => targets.includes(item.label));
+
+    if (!target) {
+      throw new Error('Menu item not found');
+    }
+
+    const targetWindow = BrowserWindow.getFocusedWindow() || BrowserWindow.getAllWindows()[0];
```
markbrockhoff: 🔧 In case you know it already, using `||` to define a fallback in case the first statement is null or undefined works but there's a safer way:
You can use the `??` (nullish coalescing operator) instead. The difference between it and `||` is that `||` will also trigger for e.g. the value 0 or an empty string. While `??` really only takes effect when the value is null or undefined. In this case it doesn't make a difference but it's just more explicit to convey what you want to archive. Also safer when dealing with primitives like strings or numbers :)


```suggestion
    const targetWindow = BrowserWindow.getFocusedWindow() ?? BrowserWindow.getAllWindows()[0];
```

markbrockhoff (commented):
LGTM overall, just some small points for potential improvements :)

e2e-tests/specs/regression/menuBar.spec.ts
zhannach: Good point, I completely overlooked defining the protocol for 1:1 conversation.

screendriver (approved):

PR #9686, opened 2026-07-02
[WPB-26754] e2e-tests folder organisation

lizaveta-korshuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26754" title="WPB-26754" target="_blank">StoryWPB-26754</a>  [Desktop/QA] Write Logout tests
  </td></table>
  <br />

### DIFF
--- a/e2e-tests/actions/createApp.ts
+++ b/e2e-tests/actions/createApp.ts
@@ -17,14 +17,9 @@
  *
  */
 
-import {_electron as electron, ElectronApplication, Page} from '@playwright/test';
+import {_electron as electron} from '@playwright/test';
 
-export type App = ElectronApplication & {
-  /* The playwright page for the main electron window wrapping the webapp */
-  wrapper: Page;
-  /* The playwright page for the currently shown webapp */
-  page: Page;
-};
+import {App} from '../types';
 
 export const createApp = async (options: {env?: string; lang?: string; dataDir: string}): Promise<App> => {
   if (!options.env) {
--- a/e2e-tests/actions/createGroup.ts
+++ b/e2e-tests/actions/createGroup.ts
@@ -17,11 +17,9 @@
  *
  */
 
-import {User} from './createUser';
-
 import {Page} from '../fixtures';
-import {conversationsList} from '../poms/webapp/conversationList.page';
-import {groupCreationModal} from '../poms/webapp/groupCreation.modal';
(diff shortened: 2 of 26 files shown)


screendriver (changes requested):
Thanks for the cleanup. I like the direction of making the e2e test structure easier to navigate.

One thing I would like to avoid, is introducing central index.ts barrel files and a dedicated types module as the main organisation mechanism.

For me, this makes the dependency graph less explicit. When I read an import like from '../../actions' or from '../../poms/webapp', I need to jump into another file to understand where something actually comes from. With direct imports, the module boundary stays visible at the call site.

I would also prefer to keep types next to the modules they describe. For example, App belongs close to createApp, User belongs close to user creation, and Team belongs close to team creation. A central types module tends to become a bucket for unrelated concepts over time, and then the ownership of those types becomes less clear.

Could we keep the direct imports and colocate the types with their owning modules instead?

markbrockhoff:
> Thanks for the cleanup. I like the direction of making the e2e test structure easier to navigate.
> 
> One thing I would like to avoid, is introducing central index.ts barrel files and a dedicated types module as the main organisation mechanism.
> 
> For me, this makes the dependency graph less explicit. When I read an import like from '../../actions' or from '../../poms/webapp', I need to jump into another file to understand where something actually comes from. With direct imports, the module boundary stays visible at the call site.
> 
> I would also prefer to keep types next to the modules they describe. For example, App belongs close to createApp, User belongs close to user creation, and Team belongs close to team creation. A central types module tends to become a bucket for unrelated concepts over time, and then the ownership of those types becomes less clear.
> 
> Could we keep the direct imports and colocate the types with their owning modules instead?

Yes, that's also the opinion I share.
I also feel like the solution suggested by this PR isn't really solving the root cause. Because I agree that working with imports in this repository is currently causing some friction. E.g. imports are not always recognized right away so they need to manually be added instead of the IDE adding them for us. Also the eslint setup enforces strict import ordering but fails to auto fix it. E.g. I've experienced myself that I have to execute an eslint fix multiple times just to get the imports into a shape which will pass CI.
So how about we try to target those issues first in order to improve developer experience? :)

lizaveta-korshuk:
@markbrockhoff @screendriver okay, I agree, for now I will close this PR

PR #9687, opened 2026-07-07
[WPB-26754] Logout tests

lizaveta-korshuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26754" title="WPB-26754" target="_blank">StoryWPB-26754</a>  [Desktop/QA] Write Logout tests
  </td></table>
  <br />


e2e-tests/poms/app/logoutPopup.page.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import {App} from '../../actions/createApp';
+
+export const logoutPopup = (app: App) => {
+  return {
+    getTitle: () => app.page.getByTestId('status-modal-title'),
+    getClearDataCheckbox: () => app.page.getByTestId('modal-option-checkbox'),
+    getCancelButton: () => app.page.getByTestId('do-secondary'),
+    getLogoutButton: () => app.page.getByTestId('do-action'),
```
zhannach: Can we convert these to direct variables, to keep our page object style consistent (matching loginPage).

```
title: app.page.getByTestId('status-modal-title'),
clearDataCheckbox: app.page.getByTestId('modal-option-checkbox'),
cancelButton: app.page.getByTestId('do-secondary'),
logoutButton: app.page.getByTestId('do-action')
```

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  await test.step("User clicks 'Log out' option", async () => {
+    const logoutSidebarButton = accountsSidebar(app).logoutButton;
+    await logoutSidebarButton.click();
+    const popupTitle = logoutPopup(app).getTitle();
+    await expect(popupTitle).toBeVisible();
+    const popupTitleText = await popupTitle.textContent();
+    await expect(popupTitleText).toBe('Clear Data?');
+    const logoutPopupButton = logoutPopup(app).getLogoutButton();
+    await expect(logoutPopupButton).toBeVisible();
+  });
```
zhannach, line 58: Since these locators are only used once within this step, we can simplify this by chaining the actions directly. We can also use Playwright's native toHaveText matcher, which handles auto-waiting perfectly under the hood.

What do you think about streamlining it like this?

```
await test.step("User clicks 'Log out' option", async () => {
  await accountsSidebar(app).logoutButton.click();
 
  await expect(logoutPopup(app).getTitle()).toHaveText('Clear Data?');
  await expect(logoutPopup(app).getLogoutButton()).toBeVisible();
});
```

Also I noticed that this and the next are 2 parts of the same user action.

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  await test.step('User logs in into Account A again', async () => {
+    await loginUser(app.page, userA);
+    await expect(conversationsSidebar(app.page).userAvatar).toContainText(userA.initials);
+    await conversationsList(app.page).getConversation(userB.fullName, {protocol: 'mls'}).open();
+    await expect(conversation(app.page).getMessage({content: 'Test message'})).toBeVisible();
+  });
+
+  await test.step('User A opens the conversation with User B', async () => {
+    await conversationsList(app.page).getConversation(userB.fullName, {protocol: 'mls'}).open();
+    await expect(conversation(app.page).getMessage({content: 'Test message'})).toBeVisible();
```
zhannach, line 94: Seems like duplication of 77-78 lines

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    const userDataDir = await app.evaluate(async ({app: electronApp}) => {
+      return electronApp.getPath('userData');
+    });
+    const sessionStorePath = path.join(userDataDir, 'Default', 'Local Storage');
+
+    await accountsSidebar(app).sidebar.click(); //workaround step to remove the context menu
+    await accountsSidebar(app).logOut(0);
+    const clearDataCheckbox = logoutPopup(app).getClearDataCheckbox();
+    await expect(clearDataCheckbox).toBeVisible();
+    await clearDataCheckbox.click({force: true});
```
zhannach: Please avoid using the { force: true } option here, as it bypasses Playwright’s native actionability checks (like visibility and interceptability). Instead, you can click on its label, just like we handle checkboxes in our web tests

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    await test.step("User right-clicks on the Account B's avatar on the sidebar with accounts' avatars and clicks 'Log out' option", async () => {
+      await accountsSidebar(app).logOut(1);
+      const userBAccount = accountsSidebar(app).getAccount(userB);
+      await expect(userBAccount.activeBorder).toBeVisible();
+      const userAAccount = accountsSidebar(app).getAccount(userA);
+      await expect(userAAccount.activeBorder).toBeHidden();
+    });
+
+    await test.step("User clicks 'Log Out' button on the popup", async () => {
+      app.page = app.windows()[2];
```
zhannach: I assume for this we have accountsSidebar(app).switchAccount(1)

e2e-tests/specs/criticalFlow/logout.spec.ts
lizaveta-korshuk: Thank you for the note, I will correct this. 


> this and the next are 2 parts of the same user action

Could you please clarify what you mean by this?

e2e-tests/specs/criticalFlow/logout.spec.ts
lizaveta-korshuk: switchAccount has an additional click inside, and I only need to change the page, because the account is switched automatically in the previous step

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    await expect(conversation(app.page).getMessage({content: 'Test message'})).toBeVisible();
+  });
+
+  await test.step('User A logs out checking the Clear Data checkbox', async () => {
+    const userDataDir = await app.evaluate(async ({app: electronApp}) => {
+      return electronApp.getPath('userData');
+    });
+    const sessionStorePath = path.join(userDataDir, 'Default', 'Local Storage');
+
+    await accountsSidebar(app).sidebar.click(); //workaround step to remove the context menu
```
zhannach: Could you link the Jira bug ticket to this comment?

lizaveta-korshuk: Please check now

zhannach (dismissed):

package.json
```diff
@@ -23,6 +23,7 @@
     "lodash": "4.18.1",
     "logdown": "3.3.1",
     "minimist": "1.2.8",
+    "open": "^11.0.0",
```
zskhan: I noticed `open` is added to package.json, but I couldn't find any imports or usages of it in this PR. Is this dependency actually needed, or was it added unintentionally?

zskhan (changes requested):

e2e-tests/poms/webapp/logoutModal.page.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import {App} from '../../actions/createApp';
+
+export const logoutPopup = (app: App) => ({
+  title: app.page.getByTestId('status-modal-title'),
+  clearDataCheckbox: app.page.getByTestId('modal-option-checkbox'),
+  cancelButton: app.page.getByTestId('do-secondary'),
+  logoutButton: app.page.getByTestId('do-action'),
+});
```
arjita-mitra, line 28: ### Modal Dismissal Not Tested

The "Cancel" button on the logout popup is created but never tested.
ESC key dismissal not tested.

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    await test.step("User right-clicks on the Account B's avatar on the sidebar with accounts' avatars and clicks 'Log out' option", async () => {
+      await accountsSidebar(app).logOut(1);
+      const userBAccount = accountsSidebar(app).getAccount(userB);
+      const userAAccount = accountsSidebar(app).getAccount(userA);
+      await expect(userBAccount.activeBorder).toBeVisible();
+      await expect(userAAccount.activeBorder).toBeHidden();
+    });
+
+    await test.step("User clicks 'Log Out' button on the popup", async () => {
+      app.page = app.windows()[2];
```
arjita-mitra: Sensing a possible race condition here

app.page = app.windows()[2] relies on window order
If windows are created/destroyed in different order, this breaks. Shall we Use app.waitForEvent('window') or more stable window tracking.

e2e-tests/actions/loginUser.ts
```diff
@@ -25,13 +25,24 @@ import {conversationsSidebar} from '../poms/webapp/conversationsSidebar.page';
...
+/* Visit the sso page and execute the login for the user */
+export const loginUser = async (page: Page, user: User) => {
+  await fillLoginCredentials(page, user);
+  await conversationsSidebar(page).userAvatar.waitFor({state: 'visible', timeout: LOGIN_TIMEOUT});
+};
+
+export const loginUserAfterDataCleanup = async (page: Page, user: User) => {
+  await fillLoginCredentials(page, user);
+  const historyConfirmButton = loginPage(page).historyConfirmButton;
+  await historyConfirmButton.click();
```
arjita-mitra, line 45: clicking historyConfirmButton without checking if it's visible assumes the history confirmation modal always appears. Should run expect().toBeVisible() first.

package.json
lizaveta-korshuk: I was meant for testing deep links, it will be needed later, but I could remove it for now or leave it for future
It's added anyway but not explicitly

e2e-tests/poms/webapp/logoutModal.page.ts
lizaveta-korshuk: This logic is supposed to be tested on web, cancelButton has been added to have full functionality of the pop up available for future need

e2e-tests/specs/criticalFlow/logout.spec.ts
lizaveta-korshuk: The idea is that windows are already created and are not destroyed.
We are checking that (while being on the account A) clicking the logout option in the account's B context menu will trigger the focus on the account B. But then we still need to manually assign app.page to account B which has index 2 because it doesn't happen automatically

lizaveta-korshuk: Alternatively we could find a not active window, e.g. like this

`const userBWindow = app.windows().find(win => win !== app.page && win.url().startsWith('http'));`

`if (userBWindow) {`
       ` app.page = userBWindow;`
     ` }`

arjita-mitra: this is much better i think it finds the window by logic, not position and works regardless of window count or creation order. 

Consider extracting this into a helper function since it might be useful in other tests.

```
export const switchToNonActiveWindow = async (app: App) => {
  const inactiveWindow = app.windows().find(
    win => win !== app.page && win.url().startsWith('http')
  );
  
  if (!inactiveWindow) {
    throw new Error('No inactive window found');
  }
  
  app.page = inactiveWindow;
};
```

package.json
zskhan: Great then lets remove it for now and add it again when needed.

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    await test.step("User right-clicks on the Account B's avatar on the sidebar with accounts' avatars and clicks 'Log out' option", async () => {
+      await accountsSidebar(app).logOut(1);
+      const userBAccount = accountsSidebar(app).getAccount(userB);
+      const userAAccount = accountsSidebar(app).getAccount(userA);
+      await expect(userBAccount.activeBorder).toBeVisible();
+      await expect(userAAccount.activeBorder).toBeHidden();
+    });
+
+    await test.step("User clicks 'Log Out' button on the popup", async () => {
+      accountsSidebar(app).switchToNonActiveWindow();
```
markbrockhoff: Could you please elaborate why manually switching to a non active window of the app is needed? If the app needs to display a popup I'd expect the app to move to it automatically. If that's what you're trying to observe, please just use the existing `watchActiveAccount()` action. You can see it in action within the notification tests

lizaveta-korshuk: Thank you, I've used watchActiveAccount instead and removed switchToNonActiveWindow logic.

e2e-tests/actions/loginUser.ts
markbrockhoff: Actions within playwright already include standard checks like e.g. that an element is visible and stable. So running `toBeVisible()` before calling `.click()` on it is simply redundant. If the element isn't there or hidden playwright will show you effectively the same error.

e2e-tests/specs/criticalFlow/logout.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    await expect(accountsSidebar(app).logoutButton).toBeVisible();
+    await expect(accountsSidebar(app).removeAccountButton).toBeVisible();
+  });
+
+  await test.step("User clicks 'Log out' option", async () => {
+    await accountsSidebar(app).logoutButton.click();
+    await expect(logoutPopup(app).title).toHaveText('Clear Data?');
+    await expect(logoutPopup(app).logoutButton).toBeVisible();
+    await expect(logoutPopup(app).cancelButton).toBeVisible();
+    await expect(logoutPopup(app).closeButton).toBeVisible();
```
markbrockhoff: Isn't it maybe a bit overkill to assert visibility for every element within the dialog? Wouldn't it be sufficient to verify the dialogs title is correct? Same for the assertions for it not to be shown.

If you want the extra safety the whole modal isn't there you could turn it into an enhanced locator and the assert on the modal itself instead of just its contents e.g.:

```ts
await expect(logoutModal(app)).toContainText('Clear Data?');
await expect(logoutModal(app)).toBeHidden();
```

e2e-tests/poms/app/logoutPopup.page.ts
```diff
@@ -0,0 +1,28 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {App} from '../../actions/createApp';
+
+export const logoutPopup = (app: App) => ({
```
markbrockhoff: 🎨  Just for consistency: Could you please not call this "popup"? There are already the terms "modal" and "dialog" floating around the codebase, adding a more just sparks confusion.

I also don't seen the need for this POM to accept the whole app, especially as it's part of the webapp. Could you please move it to the webapp poms instead?

markbrockhoff (changes requested):

e2e-tests/specs/criticalFlow/logout.spec.ts
lizaveta-korshuk: I will keep only the title for now then, thank you!

markbrockhoff (approved):

zskhan (approved):

PR #9689, opened 2026-07-09
test: write calling regression tests [WPB-26928]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26928" title="WPB-26928" target="_blank">StoryWPB-26928</a>  [Desktop/QA] Write calling regression tests
  </td></table>
  <br />
 

This PR introduces two new test cases to verify core functionality and error handling for the calling feature.

TC-11291: Verify call window maximization in 1:1 and group calls
Ensures that the call window correctly expands to full screen/maximized state during both one-on-one and group calling scenarios.

TC-11297: Verify call establishment fails without required permissions
Validates the application's error handling and permission guardrails, ensuring a call cannot be initiated if camera/microphone permissions are denied.

### DIFF
--- a/e2e-tests/specs/regression/calling.spec.ts
+++ b/e2e-tests/specs/regression/calling.spec.ts
@@ -25,7 +25,7 @@ import {callCell} from '../../poms/webapp/callCell.page';
 import {conversation} from '../../poms/webapp/conversation.page';
 import {conversationsList} from '../../poms/webapp/conversationList.page';
 
-test.describe('Calling', () => {
+test.describe('Calling - Feature Functionality', () => {
   test(
     'Verify call window maximization in 1:1 and the group call',
     {tag: ['@TC-11291', '@regression']},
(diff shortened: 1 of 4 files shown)


e2e-tests/specs/regression/calling.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      const userBPage = await createPage();
+
+      await Promise.all([loginUser(userAPage, userA), loginUser(userBPage, userB)]);
+      await connectWithUser(userAPage, userB);
+
+      await conversationsList(userAPage).getConversation(userB.fullName, {protocol: 'mls'}).open();
+      await conversation(userAPage).startCallButton.click();
+
+      await expect(app.page.getByText('No camera access')).toBeVisible();
+      await expect(conversation(userAPage).systemMessages.filter({hasText: 'You called'})).toBeVisible();
```
markbrockhoff: Is it actually intended that the system message is shown? Is the call even started when permissions are missing? Or could this be a bug?

zhannach: The call hasn't started, but the system message is displayed (only visible to the caller). We have the same behavior in production.

markbrockhoff: Ok, but let's maybe raise this and confirm, otherwise we might be testing broken behavior. (Or we just remove this assertion since IMHO it's not really relevant for this test case unless it's explicitly specified)

markbrockhoff (dismissed):

markbrockhoff (approved):

screendriver (approved):

PR #9722, opened 2026-08-18
feat(windows): add native MSI installer (WPB-5221)

adamlow-wire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5221" title="WPB-5221" target="_blank">EpicWPB-5221</a>  [Desktop] Windows installer
  </td></table>
  <br />
 

## Summary

Adds a native, per-machine Windows MSI installer for enterprise deployment through MDM and software-management systems.

The existing Squirrel EXE path remains the primary distribution channel and is built, signed, verified, archived, deployed, and promoted alongside the MSI. MSI installations deliberately do not run the application-managed Squirrel updater; enterprises update them by deploying a newer MSI with the same permanent upgrade code.

Jira: [WPB-5221](https://wearezeta.atlassian.net/browse/WPB-5221)

## What changed

- adds an `electron-builder` MSI target with stable product-family upgrade codes
- installs per-machine under Program Files and supports quiet `msiexec` deployment
- preserves desktop and Start menu shortcuts plus the `wire:` protocol registration
- supports a managed, machine-wide `WIRE_WEBAPP_URL` with validated HTTPS configuration
- adds branded installer UI and an embedded Wire banner
- prevents the Squirrel updater from starting when an MSI installation has no `Update.exe`
- keeps Squirrel release semantics intact and requires the complete existing Squirrel artifact set before Windows deployment
- builds the MSI before Squirrel mutates the packaged application, with a guard rejecting `Squirrel.exe` in MSI inputs
- signs and verifies the versioned MSI in the Windows release pipeline
- deploys MSI and Squirrel artifacts to separate S3 paths and generates URLs from their actual paths
- documents MDM deployment, upgrades, detection, migration risks, and release acceptance criteria

## Validation completed

- build/deployment unit tests: 32 passing
- updater and Windows MSI configuration compatibility tests: 8 passing
- TypeScript test compilation passes
- ESLint, Prettier, Groovy syntax, `git diff --check`, and per-commit patch checks pass
- clean GitHub [Build and Test](https://github.com/wireapp/wire-desktop/actions/runs/32820974503), [lint](https://github.com/wireapp/wire-desktop/actions/runs/32820974515), and [Semgrep](https://semgrep.dev/orgs/wire_com/projects/521372/scans/216047279) checks pass
- latest `dev` (`e1ba98c50dce28b26b05466169fbdf941f0285f3`) has no merge conflicts in a non-mutating merge-tree check
- earlier local unsigned x64 validation confirmed native MSI creation, expected MSI tables and branding, and quiet administrative extraction

## Release gates still required

These require Wire's controlled Windows Jenkins/signing environment and representative Windows/MDM devices; no safe feature-branch Jenkins trigger is exposed from this development environment.

- run the production-equivalent Windows branch build and confirm a complete signed Squirrel and MSI artifact set
- verify the final MSI payload does not contain `Squirrel.exe`, while the Squirrel package retains its expected updater payload
- verify the MSI and packaged executables with `signtool.exe verify /pa /all /tw`
- run ICE/WiX validation on the signed release artifact
- test fresh install, repair, uninstall, and N-to-N+1 upgrade as SYSTEM on supported Windows versions
- test `WIRE_WEBAPP_URL` install, retention, replacement, and clearing under managed deployment
- test migration from an existing per-user Squirrel installation
- validate assignment, detection, rollback, and exit-code handling in the target MDM tenant
- update the external Windows smoke-test job and download-page consumer before MSI promotion

## Security

- no new network endpoints, credentials, or privilege-escalation mechanism
- installer scope is intentionally per-machine and therefore requires administrative deployment authority
- the managed endpoint accepts credential-free HTTPS URLs only
- production signing uses SHA-256, timestamping, fail-fast signing, and verification of all signatures

[WPB-5221]: https://wearezeta.atlassian.net/browse/WPB-5221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bin/build-tools/lib/build-windows-msi.ts
```diff
@@ -0,0 +1,253 @@
...
+  );
+
+  const electronBuilderOsCondition =
+    /<Condition Message="Windows 7 and above is required"><!\[CDATA\[Installed OR VersionNT >= 601\]\]><\/Condition>/;
+  if (!electronBuilderOsCondition.test(customizedProject)) {
+    throw new Error('Could not find the operating system condition in the generated MSI project.');
+  }
+  customizedProject = customizedProject.replace(
+    electronBuilderOsCondition,
+    '<Condition Message="Windows 10 or above is required"><![CDATA[Installed OR VersionNT >= 1000]]></Condition>',
```
screendriver: This prevents installation on Windows 10. Windows Installer reports `VersionNT = 603` on Windows 10 for compatibility so `VersionNT >= 1000` evaluates to false. We need a Windows 10+ launch condition here, for example based on `WindowsBuild`.

See also the official Microsoft documentation: https://learn.microsoft.com/en-us/troubleshoot/windows-client/application-management/versionnt-value-for-windows-10-server

jenkins/windows.groovy
```diff
@@ -56,21 +56,42 @@ node('windows') {
...
+      echo 'Skipping application signing for non-production build (nightly/custom).'
+    }
+  }
+
+  stage('Build MSI') {
     try {
       withEnv(["PATH+NODE=${NODE}", 'npm_config_target_arch=x64']) {
         if (production || custom) {
-          bat 'yarn build:win:installer'
+          bat 'yarn build:win:msi -m'
```
screendriver: I think we should have the MSI to be created alongside the currently created `.exe` installer. This replaces the existing _Squirrel_ production build in `Wrapper_Windows_Production` and there is no separate production _Squirrel_ job in Jenkins, so after this change we would stop producing new `Wire-Setup.exe` / `.nupkg` releases.

bin/build-tools/lib/build-windows-msi.ts
```diff
@@ -0,0 +1,253 @@
...
+    buildVersion: commonConfig.version.replace(/-.*$/, ''),
+    copyright: commonConfig.copyright,
+    directories: {
+      buildResources: commonConfig.electronDirectory,
+      output: commonConfig.distDir,
+    },
+    extraMetadata: {
+      author: {name: windowsMsiConfig.manufacturer},
+    },
+    msi: {
```
screendriver, line 187: I may be missing this, but I don't see a way for the MSI deployment to configure the backend URL. The according ticket explicitly calls this out so that enterprise deployments don't need to create `init.json` or set an environment variable manually.

Is this handled somewhere else or is the MSI configuration for the backend URL still missing?

bin/build-tools/lib/build-windows-msi.ts
adamlow-wire: Good catch. Fixed in [736e10b1](https://github.com/wireapp/wire-desktop/commit/736e10b1) with the condition “Installed OR (VersionNT >= 603 AND WindowsBuild >= 10240)”, plus a regression test. Apparently Windows 10 is 6.3 when Windows Installer asks.

jenkins/windows.groovy
adamlow-wire: Agreed. Fixed in [a3be121c](https://github.com/wireapp/wire-desktop/commit/a3be121c): the job now builds Squirrel and MSI from the same signed app, signs/verifies/archives both, and deploys them to separate paths. Squirrel packaging failures are fatal too, so the MSI cannot politely hide a broken existing release.

bin/build-tools/lib/build-windows-msi.ts
adamlow-wire: This was missing. Fixed in [b595c110](https://github.com/wireapp/wire-desktop/commit/b595c110): the MSI accepts WIRE_WEBAPP_URL, stores it machine-wide, retains it across upgrades, and supports replacement or clearing. The app gives it precedence over --env and init.json, and validates it as credential-free HTTPS.

jenkins/deployment.groovy
```diff
@@ -389,7 +422,7 @@ node('built-in') {
   // ------------------------------------------------------------------------
   // Windows-specific stage: Update RELEASES file for Squirrel auto-updates
   // ------------------------------------------------------------------------
-  if (projectName.contains('Windows')) {
+  if (projectName.contains('Windows') && !isWindowsMsi) {
```
screendriver: `isWindowsMsi` is no longer defined (`hasWindowsMsi` and `hasWindowsSquirrel` are now used above) so this should fail when the Jenkins deployment reaches this stage. Since a build can now intentionally contain both installer families I think this should be based on `hasWindowsSquirrel` so we still update the Squirrel `RELEASES` feed whenever those artifacts are present.

jenkins/deployment.groovy
```diff
@@ -85,44 +93,69 @@ node('built-in') {
         // 1) Windows S3 Upload
         // -----------------------------
         env.S3_PATH = ''
+        env.MSI_S3_PATH = ''
         def AWS_ACCESS_KEY_CREDENTIALS_ID = ''
         def AWS_SECRET_CREDENTIALS_ID = ''
 
         if (params.Release == 'Production') {
           env.S3_PATH = 'win/prod'
+          env.MSI_S3_PATH = 'win/msi/prod'
```
screendriver, line 112: The MSI is uploaded to `MSI_S3_PATH` here, but the later [Generate & Store Presigned URLs stage](https://github.com/wireapp/wire-desktop/blob/b595c11043887555e90a82448d4ea28c2ee18098/jenkins/deployment.groovy#L340-L343) iterates all `wrap/dist` artifacts and [always generates URLs](https://github.com/wireapp/wire-desktop/blob/b595c11043887555e90a82448d4ea28c2ee18098/jenkins/deployment.groovy#L398-L405) using `env.S3_PATH`. That means the MSI URL will point to `win/prod/...` instead of `win/msi/prod/...`

Can we select the S3 path based on the artifact there?

jenkins/deployment.groovy
adamlow-wire: Quite right — the ghost of `isWindowsMsi` was still rattling around. Fixed in [5e987625](https://github.com/wireapp/wire-desktop/commit/5e987625): feed promotion now keys off `hasWindowsSquirrel`. [255bdd54](https://github.com/wireapp/wire-desktop/commit/255bdd54) also makes a complete Squirrel set mandatory for Windows deployment. Regression coverage included.

jenkins/deployment.groovy
adamlow-wire: Quite right — we uploaded the MSI to one cupboard and handed out directions to another. Fixed in [3588bc9a](https://github.com/wireapp/wire-desktop/commit/3588bc9a): both presigned-URL branches now use `MSI_S3_PATH` for `.msi`, while every other artifact stays on `S3_PATH`. Regression coverage included.

bin/build-tools/lib/build-windows-msi.ts
```diff
@@ -0,0 +1,272 @@
...
+  };
+
+  return {builderConfig, windowsMsiConfig};
+}
+
+export async function validateWindowsMsiAppDirectory(appDirectory: string, executableName: string): Promise<void> {
+  if (!(await fs.pathExists(path.join(appDirectory, `${executableName}.exe`)))) {
+    throw new Error(`Packaged Windows application not found in "${appDirectory}". Run the Windows build first.`);
+  }
+  if (await fs.pathExists(path.join(appDirectory, 'Squirrel.exe'))) {
```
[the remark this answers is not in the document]
adamlow-wire: /fp Build-time packaging only: appDirectory is derived from repository/CI build configuration, and this helper only checks that expected package inputs exist. It is not reachable from runtime or untrusted user input and does not read or expose file contents.

bin/build-tools/lib/build-windows-msi.ts
```diff
@@ -0,0 +1,272 @@
...
+      signtoolOptions: manualSign ? {sign: async () => undefined} : undefined,
+      target: ['msi'],
+    },
+  };
+
+  return {builderConfig, windowsMsiConfig};
+}
+
+export async function validateWindowsMsiAppDirectory(appDirectory: string, executableName: string): Promise<void> {
+  if (!(await fs.pathExists(path.join(appDirectory, `${executableName}.exe`)))) {
```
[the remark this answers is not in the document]
adamlow-wire: /fp Build-time packaging only: executableName comes from repository/CI build configuration, and this helper only checks that the expected packaged executable exists. It is not reachable from runtime or untrusted user input and does not read or expose file contents.

screendriver (approved):
LGTM 👍

